PR #25, opened 2026-04-12
Full repo review

bsaffel:
Full codebase review via /code-review:code-review skill.


bsaffel:
### Code review

Found 5 issues:

1. Stale CLI commands in MCP error messages — five error strings tell users to run `moneybin extract ofx` / `moneybin extract w2`, but those commands don't exist; the correct commands since the CLI restructure are `moneybin data extract ofx` / `moneybin data extract w2`

https://github.com/bsaffel/moneybin/blob/645844159a56add488df2c3733f00b4ab637924e/src/moneybin/mcp/tools.py#L149-L153

2. `toggle_category` always reports success — `result.fetchone` is a bound method (never `None`), so the condition on line 271 is always `True` and the "not found" branch is unreachable; toggling a non-existent category silently claims success

https://github.com/bsaffel/moneybin/blob/645844159a56add488df2c3733f00b4ab637924e/src/moneybin/mcp/write_tools.py#L269-L275

3. Path traversal not validated in `import_file` MCP tool — `file_path` from the MCP client is passed directly to `Path(file_path)` with no check against `..` or absolute paths outside expected directories ([CLAUDE.md](https://github.com/bsaffel/moneybin/blob/645844159a56add488df2c3733f00b4ab637924e/CLAUDE.md) says "Validate paths against traversal (`..`, absolute paths).")

https://github.com/bsaffel/moneybin/blob/645844159a56add488df2c3733f00b4ab637924e/src/moneybin/mcp/write_tools.py#L66-L83

4. Import date range reports global history, not current file — `SELECT MIN/MAX FROM raw.ofx_transactions` has no `WHERE source_file = ?` filter, so `ImportResult.date_range` grows to the all-time historical range after the second import

https://github.com/bsaffel/moneybin/blob/645844159a56add488df2c3733f00b4ab637924e/src/moneybin/services/import_service.py#L143-L153

5. Merchant matches silently override rules regardless of priority — `apply_deterministic_categorization` calls `apply_merchant_categories` first, then `apply_rules` on remaining uncategorized transactions; the `apply_rules` docstring promises "first matching rule wins" in priority order, but any merchant match pre-empts all rules unconditionally

https://github.com/bsaffel/moneybin/blob/645844159a56add488df2c3733f00b4ab637924e/src/moneybin/services/categorization_service.py#L373-L379

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

bsaffel:
All comments addressed in #26

PR #50, opened 2026-04-25
Simplify CLI startup flow and auto-init database on profile create

bsaffel:
## Summary

Eliminates two sources of startup complexity that caused bugs during profile lifecycle operations: module-level profile initialization that ran before the CLI callback could set the profile, and `setup_logging()` calling `get_settings()` which failed when no profile was set. Additionally, `profile create` now auto-initializes the encrypted database, removing the separate `db init` step for new profiles.

## Impact

- `get_settings()` and `get_current_profile()` now raise `RuntimeError` if called before `set_current_profile()` — code that relied on the implicit "default" fallback needs updating
- `setup_logging()` signature changed: removed `profile`, `cli_mode`, `config` parameters; replaced with explicit `level`, `log_format`, `log_to_file`, `log_file_path`
- `profile create` now auto-initializes the encrypted database — no separate `db init` step needed
- `clear_settings_cache()` resets profile to `None` instead of `"test"` — test fixtures must call `set_current_profile()` after clearing
- `init_db()` and `derive_key_from_passphrase()` now live in `database.py` with `secret_store` parameter for dependency injection — no more patching `SecretStore` at multiple import sites

### Config module (`config.py`)
Removed eager initialization that read user config at import time. `_current_profile` starts as `None`, set exclusively by `set_current_profile()` in `main_callback()`. Removed `create_directories()`, `_get_initial_profile()`, `create_dirs` field, and `get_logging_config()` — all were either side-effect-laden or unused.

### Logging decoupling (`logging/config.py`, `observability.py`)
`setup_logging()` no longer calls `get_settings()`. All config values are passed explicitly by `setup_observability()`, which resolves them from settings when a profile is available or uses defaults (console-only, INFO) when not. Added `_ConsoleNoiseFilter` to suppress noisy SQLMesh INFO messages from console output. Fixed atexit handler to use `get_database_if_initialized()` instead of importing the private `_database_instance`.

### CLI callback (`cli/main.py`)
Restructured `main_callback()` with explicit `needs_profile` flag. Profile commands (`ctx.invoked_subcommand == "profile"`) skip all profile resolution — no `ensure_default_profile()`, no `set_current_profile()`, no directory validation.

### Database and DB init (`database.py`, `db.py`)
Changed `Database.__init__` mkdir from `parents=True` to `parents=False` so deleted profile directories can't be silently recreated. Moved `_derive_key_from_passphrase` and DB init logic from CLI `db.py` to `database.py` as `derive_key_from_passphrase()` and `init_db()` with `secret_store` parameter for dependency injection — callers pass their own `SecretStore` instance, eliminating multi-site monkeypatching in tests.

### Profile auto-init (`profile_service.py`, `profile.py`)
`ProfileService.create()` now calls `_init_database()` to auto-create the encrypted database at profile creation time. E2E test infrastructure updated with `MemoryKeyring` backend for subprocess key round-trip isolation.

### E2E test reliability (`conftest.py`)
Added `start_new_session=True` to subprocess runner so `getpass` (used by `hide_input=True` prompts) falls back to reading stdin instead of `/dev/tty`. Without this, piped `input_text` was ignored and passphrase tests hung waiting for terminal input.

### Architecture documentation
Added `docs/tech/cli-startup-flow.md` documenting the full CLI startup sequence with Mermaid diagrams, invariants, and the simplifications applied.

## Testing

- 1082 unit tests pass
- 23 integration tests pass
- 97 E2E tests pass (including new `test_profile_create_initializes_database` and `test_profile_create_runs_migrations`)
- 0 pyright type-check errors

## Notes

- The `src/moneybin/services/metrics_service.py` and `src/moneybin/services/synthetic_service.py` files in the working directory are linter-generated orphans not imported by anything — intentionally excluded from this commit.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -84,6 +84,10 @@ pyrightconfig.json
 
 # Claude Code local settings (personal, not shared)
 .claude/settings.local.json
+.claude/scheduled_tasks.lock
+
+# Stray local config (real config lives under <base>/config.yaml)
+/config.yaml
 
 # MCP config (contains local paths and credentials)
 .mcp.json
--- a/docs/tech/cli-startup-flow.md
+++ b/docs/tech/cli-startup-flow.md
@@ -1,0 +1,283 @@
+# CLI Startup Flow
+
+How a MoneyBin CLI command goes from `moneybin <cmd>` to executing user code, what happens at each stage, and where things can go wrong.
+
+## Overview
+
+```mermaid
+flowchart TD
+    A["moneybin &lt;cmd&gt; [--profile X] [-v]"] --> B[Module-level init]
+    B --> C[main_callback]
+    C --> D{Is subcommand<br/>'profile'?}
+    D -->|yes| E[Skip profile resolution]
+    D -->|no| F[Resolve profile name]
+    F --> G[set_current_profile]
+    G --> H[Validate profile dir exists]
+    E --> I[setup_observability]
+    H --> I
+    I --> J[Subcommand executes]
+    J --> K[Process exit]
+    K --> L[atexit: flush metrics]
+```
+
+## Stage 1: Module-Level Initialization
+
+When `moneybin.config` is first imported, a module-level variable is initialized:
+
+```python
+_current_profile: str | None = None
+```
+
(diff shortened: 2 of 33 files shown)


bsaffel:
@claude if this pr is good can you mark it approved?

bsaffel:
@claude please resolve all other threads that have been resolved above

PR #60, opened 2026-04-29
Quality cleanup — error handling, auto-rule typing, observability

bsaffel:
## Summary

Bundles six related cleanup commits into one quality pass on top of `main`: a cross-cutting error-handling unification (CLI + MCP), typed auto-rule result objects, account-match observability fixes, and assorted simplifications surfaced by `/simplify`. No new features; no user-visible behavior change other than improved error messages and metric coverage.

## Impact

- **No CLI flag changes**, but `handle_database_errors` is renamed to `handle_cli_errors` — out-of-tree extensions (none known) would need to update the import.
- **`AutoRuleService` return shapes changed** from dicts to dataclasses (`AutoReviewResult`, `AutoConfirmResult`, `AutoStatsResult`). Internal-only consumers; all in-tree call sites updated.
- **`ResponseEnvelope` gained an optional `error` field**. Existing tools serialize identically when `error is None` (the key is omitted).

## Breaking Changes

- **`moneybin categorize auto-review --output json` shape change.** Previously emitted a bare JSON array of proposals; now emits a `ResponseEnvelope` object: `{summary: {total_count, returned_count, has_more, sensitivity}, data: [...proposals], actions: []}`. Proposals now live at `.data[]`. Done deliberately to honor the CLI/MCP symmetry rule in `docs/specs/mcp-architecture.md` §5 — every `--output json` returns the same envelope its MCP equivalent does, and pagination/truncation are now first-class signals rather than hidden truncation. Any pre-existing scripts indexing the JSON top-level as a list must update to `.data[]`.

### Unified user-facing error handling (ffb545b)
New `moneybin/errors.py` defines `UserError` + `classify_user_error()` — a shared classifier feeding two surface-specific adapters: `handle_cli_errors` (logs + `typer.Exit(1)`) and `handle_mcp_errors` (returns an error envelope). Drops 9 redundant per-command `try/except FileNotFoundError` blocks. ResponseEnvelope gains an `error: UserError | None` field with a matching `build_error_envelope()` helper.

### Auto-rule listings simplification (15f1c28, 2401c3f, ed600e2)
- Replaced dict returns from `AutoRuleService.review/confirm/stats` with typed dataclasses carrying their own `to_envelope()`.
- Added `--limit` / `has_more` pagination to `auto-review` and `auto-rules` listings.
- Simplified matching priority seeding to use `db.executemany`.

### Source priority seeding (f2b01ff)
Seed `app.seed_source_priority` before SQLMesh transforms run so models can reference the priority table on first invocation.

### Account-match observability (cb9a6ff)
Split the `fuzzy_warned` counter into `fuzzy_no_id` and `fuzzy_ambiguous` for clearer attribution. Broke a circular import between `import_service` and the matching subsystem.

## Testing

- 1170 unit tests pass (`pytest -m "not integration and not e2e"`).
- 104 E2E subprocess tests pass (`pytest tests/e2e -m e2e`).
- 17 new tests added: 5 for the classifier, 8 for the MCP adapter + envelope, 4 for the CLI adapter.
- Type-checked the modified set with pyright (0 errors).

## Notes

- Several deferred items (per-tool MCP migration to `@handle_mcp_errors`, FastMCP 3.x SDK migration, `_run_tier` extraction, etc.) are captured in `private/followups.md` rather than expanding this PR's scope.
- The error-handling shape (`UserError` + classifier + adapters) was deliberately built SDK-agnostic so a future move to `fastmcp` 3.x's `@handle_tool_errors` is a one-file swap, not a call-site migration.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


bsaffel:
## Review status as of `624f934`

Two automated reviews landed at 06:26–06:27 UTC, ~80 seconds after the push of `9679757`. Both reference five "open" findings, but two of them were already addressed in `9679757` — the reviewer's checkout pre-dated that commit. Posting a status snapshot so this thread isn't ambiguous.

### Already fixed at HEAD
- **`typer.Exit` caught in `handle_cli_errors`** — `cli/utils.py` now has an explicit `except typer.Exit: raise` ahead of the `Exception` catch (`9679757`).
- **`transform plan --apply` skips priority seeding** — `plan_transforms` delegates to `apply_transforms()` (which routes through `run_transforms()` → `seed_source_priority`) when `--apply` is set; the direct `ctx.plan` path is preview-only (`9679757`).
- **`FileNotFoundError` errno prefix in user message** — addressed in `624f934`.

### Deferred — captured in `private/followups.md`

These are all real findings that we've accepted but explicitly chose not to bundle into this branch:

- **`auto-review --output json` shape change.** Done deliberately to honor `docs/specs/mcp-architecture.md` §5 (CLI/MCP symmetry: every `--output json` returns the envelope its MCP equivalent does). The breaking change is documented in this PR's `## Breaking Changes` section. Pre-1.0 project, no third-party scripts in the wild.
- **`handle_mcp_errors` not yet applied to MCP tools.** Tracked as "Apply `handle_mcp_errors` across MCP tool modules" in followups. Each `except Exception: # noqa BLE001` block in the existing tools needs a per-tool audit (the DuckDB constraint-error UX would regress under a blind decorator migration). Will land as `refactor/migrate-mcp-tools-to-error-handler`, likely bundled with the FastMCP 3.x upgrade also captured in followups.
- **Service-layer dataclasses import from `moneybin.mcp.envelope`.** Real dependency-direction inversion. Tracked as "Move `to_envelope()` out of service-layer dataclasses (dependency direction)" in followups. Will fold into the MCP-tool migration PR above since both touch the same surface.

The two automated reviewers will likely re-flag the deferred items on every push until those refactors land. The deferral is intentional, not an oversight — keeping this PR focused on the unified error-handling shape rather than expanding into a multi-subsystem migration.

PR #69, opened 2026-04-30
Add 'categorize bulk' CLI + bulk-loop perf

bsaffel:
## Summary
- New `moneybin categorize bulk` CLI command with file/stdin input and `--output {table,json}`, mirroring the `categorize.bulk` MCP tool. Closes the largest CLI symmetry gap.
- Shared `BulkCategorizationItem` Pydantic model validated at every boundary; per-item validation accumulates failures into the existing `BulkCategorizationResult.error_details` envelope (no fail-fast).
- New `BulkRecordingContext` threaded through `AutoRuleService.record_categorization` eliminates per-item duplicate description / merchant / rule SELECTs. Query count drops from `~3 + 5N` to `~3 + N`.
- Auto-rule E2E test now drives the real CLI flow (`import → categorize bulk → auto-review → auto-confirm → re-import`) instead of seeding `app.proposed_rules` via raw SQL.
- New observability metrics: `categorize_bulk_items_total`, `categorize_bulk_duration_seconds`, `categorize_bulk_errors_total`.

Resolves three items in `private/followups.md`: the missing CLI command, the rule/merchant pattern cache, and the duplicate description SELECT.

## Impact
- AI agents and scripts can now bulk-categorize via CLI, not just MCP.
- No breaking changes; existing CLI commands and MCP tool semantics unchanged.

## Test plan
- [x] `uv run pytest tests/moneybin/test_categorization_service_bulk.py tests/moneybin/test_bulk_recording_context.py tests/moneybin/test_auto_rule_service_context.py -v`
- [x] `uv run pytest tests/integration/test_categorize_bulk_cli.py -v`
- [x] `uv run pytest tests/e2e/test_e2e_workflows.py::TestAutoRulePipeline -v`
- [x] `make check test`

Spec: `docs/specs/categorize-bulk.md`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -229,6 +229,7 @@ moneybin import file <path>        # Import financial data files
 moneybin transform apply           # Run SQLMesh pipeline
 moneybin transform status          # Check pipeline state
 moneybin categorize apply-rules    # Apply categorization rules
+moneybin categorize bulk           # Bulk-categorize a list of transactions
 moneybin db shell                  # Interactive SQL shell
 moneybin db ps                     # Show processes holding the database
 moneybin logs tail -f              # Follow log output
--- a/docs/guides/categorization.md
+++ b/docs/guides/categorization.md
@@ -63,6 +63,15 @@ All categorization operations support batch mode for efficient processing. These
 - `categorize.create_merchants` — create one or many merchant mappings
 - `categorize.delete_rule` — remove a rule
 
+**Via CLI** — the `categorize.bulk` tool has a CLI equivalent that accepts the same JSON shape from a file or stdin:
+
+```bash
+moneybin categorize bulk --input cats.json
+cat cats.json | moneybin categorize bulk -
+```
+
+Both surfaces share the same response envelope; pass `--output json` to get the structured result.
+
 ## Category Taxonomy
 
 MoneyBin ships with the Plaid Personal Finance Category v2 (PFCv2) taxonomy — approximately 100 default categories organized into top-level categories and subcategories.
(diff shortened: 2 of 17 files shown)


bsaffel:
Addressed the JSON-injection thread in `tests/e2e/test_e2e_workflows.py` via `json.dumps`. Notes on the four threads I'm not changing in this push:

**`auto_rule_service.py:752` and `:803` (codex)** — both premised on a "transient preload failure" where `BulkRecordingContext` is provided but `txn_rows` / `active_rules` are partially populated. That mode does not exist in this codebase: Phases 2 and 3 of `_bulk_categorize_inner` either succeed (and fully populate the context) or raise. An empty `active_rules` legitimately means "no active rules exist," not "preload failed silently." Falling back to per-item DB reads in either path would also defeat the `~3 + N` query budget that this PR is designed to enforce.

**`auto_rule_service.py:157` (claude, `merchant_row_for` O(N) scan)** — reviewer flagged as a nit ("fine for typical personal-finance scale"). Per-profile merchant counts are bounded in the hundreds; the linear scan is well below noise. Maintaining a parallel `dict[str, tuple]` would add an invariant burden across `register_new_merchant` and the `merchant_mappings` ordering rules without a measurable win. Happy to revisit if profiling shows this on a hot path.

**`categorization_service.py:377` (claude, `skipped` counter always 0)** — accurate observation. After the `validate_bulk_items` refactor, boundary failures merge into `errors`, and there's no remaining code path that increments `skipped`. The metric label, CLI exit-code branch (Req 4), and envelope field are all currently dead. This is a design call — drop the field, or add a real skip case (e.g., `transaction_id` not present in `fct_transactions`) — and I'd rather handle it as a followup than rush a choice into this PR. Filing it as a followup item.

PR #199, opened 2026-05-22
Package framework core: manifest, validators, registry

bsaffel:
## Summary

Adds `src/moneybin/packages/_framework/` — the runtime that loads, validates, and registers analysis packages declared via the `moneybin.packages` setuptools entry-points group. This is the framework foundation for the extension-contracts spec (`docs/specs/extension-contracts.md`); reference packages, the validator CLI, and report scaffolding are tracked as separate follow-on work.

What lands here:

- **`PackageManifest`** (Pydantic) parses `moneybin_package.yaml` with semver, quality-scale, lowercase-snake_case `owns_prefix`, and name-matches-prefix validation.
- **Capability validation** — `validate_writes` enforces the load-bearing rule that every `CREATE TABLE`/`CREATE VIEW` in a package's `schema/` matches a declared write glob (case-insensitive, matching DuckDB identifier semantics).
- **`discover_packages`** enumerates installed packages via `importlib.metadata`; a bad entry never takes down the framework (logged + skipped).
- **Prefix-discipline validators** check SQL writes, schema filenames, MCP tool names, and CLI subcommands against the package's `owns_prefix`.
- **Quality Scale validator** runs cumulative tier checks (Bronze → Platinum) mechanically; non-mechanical checks (signed releases, coverage thresholds) are deferred to post-launch hardening.
- **`register_package`** orchestrates validate → invoke `tools.register(mcp)` + `cli.register(app)` → add to the in-process `PackageRegistry`. Any violation raises the first error and registers nothing (no partial state).
- **`MoneyBinSettings.packages`** nested config mirrors the existing `providers` pattern.
- A **synthetic test package** under `tests/.../fixtures/test_synthetic/` is the framework's permanent regression surface, exercised end-to-end (discover → validate → register).

Also includes a small workflow-rule addition to `.claude/rules/sandboxing.md` (the `dangerouslyDisableSandbox` protocol) — separable from the framework code; flagged here so the diff isn't surprising.

Deferred to follow-on work (called out in module docstrings where relevant): reference packages (`assets`, `us_tax`), SQLMesh model-path registration, the `moneybin extension validate` CLI/MCP surface, and runtime read/network capability enforcement.

## Test plan

- [x] `make check test` green (2488 passed)
- [x] `uv run pyright src/moneybin/packages/` clean
- [x] Synthetic test package validates and registers end-to-end
- [x] Capability / prefix / quality-scale violations surface precise errors with offender info
- [x] `pyproject.toml` declares the `moneybin.packages` entry-points group (empty for now); group resolves at runtime
- [x] Public framework surface imports with no heavy deps (fastmcp/typer guarded under TYPE_CHECKING)


bsaffel:
**Re: schema DDL execution wiring (round-2 🔴)**

The review correctly notes `init_schemas()` gains `additional_files` here but `database.py` doesn't pass package schema files to it. This is a deliberate scope boundary: Plan 2 ships the framework primitives and the `init_schemas` hook; **threading discovered packages' schema files into the database bootstrap (plus path-containment validation of those files) is Plan 4 work**, landing alongside the first reference packages.

Rationale: no packages ship in this PR (the `moneybin.packages` entry-points group is empty), so there is nothing to execute in production yet. Wiring `discover_packages() → init_schemas(additional_files=...)` now would run discovery at every DB init for zero packages. The deferral is now documented in `registry.py`'s "Known deferred" block, and the path-containment check is filed as a follow-up gated on that wiring.

PR #193, opened 2026-05-20
feat: RecoveryAction + error_code taxonomy + envelope plumbing (M2D PR 2)

bsaffel:
## Summary

First implementation PR for the M2D recoverable-state milestone (spec: `docs/specs/data-recovery-contract.md`). Lands the universal envelope contract — `RecoveryAction` type, prefix-grouped error-code taxonomy, `recovery_actions` field on both `UserError` and `ResponseEnvelope`, and `build_error_envelope()` plumbing that threads the field from raise site to JSON wire.

No tool populates `recovery_actions` yet — that's PRs 9a-N (per-domain retrofit). PR 2 ships the infrastructure every other PR inherits.

## Impact

- **Backward-compatible.** Every existing `UserError(message, code=...)` call still works; the new `recovery_actions` parameter defaults to `None`.
- **Existing error codes were migrated** to the new prefixed taxonomy (per spec Req 3) — `database_locked` → `infra_database_locked`, `invalid_input` → `mutation_invalid_input`, etc. 11 downstream test assertions updated to use the new constants. If you import `classify_user_error` codes by literal string, this is a breaking change for you; switch to `error_codes.INFRA_*` / `MUTATION_*` / etc.
- **`ResponseEnvelope.to_dict()` now emits `recovery_actions`** when non-None (preserves empty list to distinguish "nothing actionable" from "field absent"). Once downstream PRs populate the field, agents will receive structured recovery actions in MCP / CLI responses.

#### `src/moneybin/error_codes.py` (new)
Prefix-grouped error-code constants:
- `import_*` — loading raw data (5 codes)
- `mutation_*` — app-state writes (4 codes)
- `audit_*` — doctor / invariant failures (4 codes)
- `refresh_*` — pipeline crashes (3 codes)
- `undo_*` — audit-log undo consumer (3 codes, used by PR 3)
- `recovery_*` — recovery tooling itself (`recovery_no_path`)
- `infra_*` — database / migrations / encryption (8 codes; absorbs pre-spec codes)
- `sync_*` — external connectors (1 code; placeholder for sync errors)

Property tests verify: every code uses a valid prefix, NAME == VALUE.upper(), values are lowercase snake_case, no duplicates.

#### `src/moneybin/errors.py` — `RecoveryAction` Pydantic v2 model
[...]

PR #202, opened 2026-05-22
RefreshResult error surfacing (M2D REC-PR6)

bsaffel:
## What

Closes the refresh-result silent-failure gap (M2D / `data-recovery-contract.md` Req 9). `refresh_run` and `moneybin refresh` previously swallowed best-effort matcher/categorizer crashes at DEBUG, so a partially-refreshed pipeline (cross-source dupes accumulating, rows left uncategorized) looked healthy.

## Changes

- **`RefreshResult`** gains `matching_error`, `categorization_error`, and a `self_heal_actions: list[SelfHealRecord]` carrier (the latter empty until the M2D self-heal safelist lands — shipped now so the result shape is stable for agents).
- **Crash surfacing.** A *real* matcher/categorizer crash now logs at ERROR and populates the matching `*_error` field; the pipeline still continues (one stage's failure never aborts the others).
- **Precondition discrimination.** A missing-view precondition (`duckdb.CatalogException`/`BinderException` — e.g. first load before SQLMesh apply built the views) stays a quiet DEBUG with no error surfaced, so a fresh database's first refresh doesn't report a false failure. This is what genuinely closes the gap; a blind DEBUG→ERROR would trade silent failure for false-positive noise. Mirrors the existing precedent in `inbox_service.py`.
- **Recovery actions.** When a step crashes, the response envelope carries structured `recovery_actions`: the targeted retry(s) (`refresh_run(steps=["match"])` / `refresh_run(steps=["categorize"])`) followed by a single `system_doctor` diagnostic. `build_envelope` gains an optional `recovery_actions` param (the success-path counterpart to `build_error_envelope`).
- **CLI parity.** Human output warns on step crashes with a retry hint; `--output json` carries the same `recovery_actions`. Best-effort crashes don't fail the command (exit 0).

## Spec corrections

Two inaccuracies in `data-recovery-contract.md` Req 9 are fixed in this PR rather than deferred:
- The `system_doctor` recovery action uses empty `arguments` — the `system_doctor` MCP tool is parameterless (`--verbose` is a CLI-only flag), so the spec's `system_doctor(verbose=True)` would emit a non-executable action.
- Req 9's "DEBUG→ERROR" now documents the precondition discrimination described above.

The spec is also promoted `draft → in-progress` (Phase 2 PR2 shipped in #193; the spec is effectively final).

## Tests

- Service unit tests: matcher/categorizer crash → `*_error` populated; precondition exception → no error; pipeline continues.
- Adapter unit tests: recovery-action ordering, `system_doctor` empty args, single doctor action, `self_heal_actions` serialization.
- `build_envelope` recovery_actions threading.
- CLI: JSON `recovery_actions` parity + human-output warning.
- Full unit suite green (975 passed).


bsaffel:
Addressed the review in 160736da:

- **🟡 Stale docstrings** — updated the MCP tool docstring + register description, the CLI command docstring, and the service module docstring. They no longer claim "only SQLMesh apply errors surface"; they now describe `matching_error`/`categorization_error` + `recovery_actions` so an agent reading the tool description knows to check those fields.
- **🟡 Contradictory categorize hint** — the follow-up hint is now gated on `result.matching_error is None`, so a matcher crash emits only the "retry match" recovery action, not a conflicting "run categorize next" hint. Added two adapter tests (suppressed-on-crash, still-fires-on-clean-match-only).
- **🟡 gsheet CLI–MCP symmetry** — resolved as functional parity rather than a new flag: `RefreshStepChoice` now documents why `--step gsheet` is intentionally omitted (the dedicated `moneybin gsheet pull` is the CLI path; a `--step gsheet` flag would be redundant), and the MCP docs now correctly list `gsheet` as a valid step (it is exposed in the MCP schema via the `RefreshStep` Literal).
- **🔵 NIT (categorization_error overload)** — added the clarifying comment on the apply-failed early return.

**Declined — 🔵 NIT (add `"transform" in requested` to the hint guard):** this would break the intended match-only follow-up. `test_refresh_step_match_without_categorize_emits_followup_hint` asserts the hint fires for `steps=["match"]` (no transform) — prompting the agent to run categorize after a standalone match pass is the designed behavior, so I left the guard as match-without-categorize.

bsaffel:
Addressed round-2 in 8b86dfb5:

- **🟡 ✅-after-⚠️ contradictory output** — the success banner is now dropped when a best-effort step crashed (`Refresh complete … (best-effort step failures above)` instead of `✅ Refresh complete`), so the success line never contradicts the warning above it.
- **🟡 quiet mode silently exits 0** — `⚠️` step-crash warnings now fire even under `--quiet` (per cli.md, `-q` suppresses informational output / `✅`, not warnings). Exit code stays 0 (best-effort steps don't fail the command, by design), but the failure is no longer silent. Added a quiet-mode test.
- **🟡 mutable `self_heal_actions` on frozen dataclass** — changed to `tuple[SelfHealRecord, ...]` (`default_factory=tuple`); the carrier is now truly immutable. `asdict` still serializes it as a JSON array.
- **🔵 `finally` "finished" on skip** — now logs "attempted".
- **🔵 BinderException untested** — both first-load tests (matcher + categorizer) are parametrized over `CatalogException` and `BinderException`.
- **🔵 `self_heal_actions==[]` contract** — asserted in `test_refresh_json_success`.
- **Open thread (P2) — match-retry recovery_action emitted alongside an apply failure** — `_step_crash_recovery_actions` now returns `[]` when `result.error` is set: the apply failure is the blocker (surfaced via `error` + the apply-failed `actions` hint), so a step retry would misdirect. Added a test.

**Still declining — open thread (categorize follow-up hint on `steps=["gsheet","match"]` without transform):** this is pre-existing hint-gating behavior, not introduced by this error-surfacing PR, and the equivalent `steps=["match"]` case is intended and tested (`test_refresh_step_match_without_categorize_emits_followup_hint` asserts the hint fires). Adding a `"transform" in requested` guard would change established, tested behavior; if that gating is wrong it deserves a focused change rather than riding on this PR.

bsaffel:
Addressed round-3 in c00c313c:

- **🔴 MUST FIX — categorize try-block scope.** Split the post-categorization auto-rule proposal read (`AutoRuleService.stats()`) out of the `categorize_pending()` try block into its own log-and-continue block. Only `categorize_pending()` now decides `categorization_error`; a crash in the informational stats read no longer falsely reports a categorization failure or misdirects recovery. Added a test (`test_refresh_auto_rule_stats_crash_is_not_a_categorization_error`).
- **🟡 JSON-mode warning suppression.** Resolved by fixing the behavior, not the comment: the `⚠️` step-crash warnings now emit *before* the output-format branch, so they fire in `--output json` too (diagnostics → stderr; JSON data stays clean on stdout, and the crash is also in the payload). Locked with an assertion in `test_refresh_matcher_crash_surfaced_in_json` (stdout parses clean JSON; warning present in combined output).
- **🟡 asdict vs model_dump.** Added a comment locking the choice: `SelfHealRecord` is a `@dataclass` like `RefreshResult` (not Pydantic), so `asdict` is the consistent serializer here; if it ever migrates to `BaseModel`, switch to `model_dump()`.
- **🔵 asymmetric `categorization_error` scoping.** Now declared once at the top of `refresh()` alongside `matching_error`, removing the latent `UnboundLocalError` risk.

Full unit suite green (2757 passed).

bsaffel:
Addressed round-4 in d831e059:

- **🔴 MUST FIX — missing test (matcher crash + apply failure co-occurrence).** Added `test_refresh_matcher_crash_preserved_when_apply_also_fails`: matcher crashes AND apply fails → `applied=False`, `error="apply boom"`, `matching_error="matcher boom"` (the matcher crash is preserved on the apply-failed return).
- **🟡 CLI 💡 retry hint fires when apply also failed.** Gated the text-mode 💡 step-retry hint on `result.error is None` — consistent with the adapter's `recovery_actions` suppression. The `⚠️` warning still shows (the crash happened); the misdirecting "re-run the failed step" hint is suppressed when the apply error (❌) is the real blocker. Added a test.
- **🟡 `recovery or None`.** The suggested `recovery if recovery else None` is behaviorally identical (`[]` → `None` either way). The concern is theoretical here: refresh always uses `build_envelope`, whose `ResponseEnvelope.error` is `None`, so there's no `error.recovery_actions` to fall through to — the suppress-vs-fallthrough contract applies only to `build_error_envelope`. Added a comment locking that reasoning; no behavior change (current behavior is tested and correct).
- **🟡 asdict vs model_dump.** Already resolved in round-3 (c00c313c) via the locking comment — `SelfHealRecord` is a `@dataclass` like `RefreshResult`; switch to `model_dump()` only if it ever migrates to Pydantic.
- **🟡 import/inbox drop the new fields.** Scope decision (option b): `refresh()` already logs crashes at **ERROR** for *all* callers (import/inbox included) — strictly better than the prior DEBUG swallow, so there's no silent-in-logs regression. Carrying `matching_error`/`categorization_error` in `BatchImportResult`/`InboxSyncResult` envelopes is a clean follow-up beyond "RefreshResult error surfacing" and is tracked for the next import/inbox envelope touch.

Full unit suite green (2759 passed).

PR #200, opened 2026-05-22
Add Invariant 10 (app.* mutation routing) contract

bsaffel:
## Summary

Phase 1, PR 1 of `app-integrity-invariant.md`. Establishes the contract the rest of Phase 1 enforces — **no service code yet**.

- **Invariant 10 — `app.* mutation routing`** appended to `architecture-shared-primitives.md`: every `app.*` mutation must emit a paired `app.audit_log` row through a `*Repo` in the same DuckDB transaction; the lint rule + doctor invariants enforce it across the later PRs. Carries a provenance note (matching Invariant 9's style) since enforcement rolls out incrementally.
- **Full-row pre-image capture** supersedes the earlier column-subset audit shape (`transaction-curation.md` Req 29, the `app.audit_log` column comments in both the spec and `src/moneybin/sql/schema/app_audit_log.sql`, and the Audit Emission Contract). This is the data foundation that keeps the Phase 2 undo consumer strictly additive.
- **ADR-012** records repository-over-decorator / over-event-sourcing and the Phase 1 / Phase 2 split.
- **AGENTS.md** Key Abstractions gains a "Protected `app.*` mutation → compose a `*Repo`" row.

## Invariant renumber (cross-spec)

Both M2D specs were written assuming `app.*` routing = Invariant 9 and recoverability = Invariant 10. Slot 9 is already held by the shipped `category_id`-FK invariant (#174), so:

- `app.*` mutation routing → **Invariant 10** (this work)
- Recoverability of mutations → **Invariant 11** (data-recovery, Phase 2)

Renumbered consistently across `app-integrity-invariant.md`, `data-recovery-contract.md`, and `INDEX.md`.

## Scope

Docs + one SQL comment block — no executable code. Subsequent PRs add the `repositories/` layer, per-table migrations, doctor invariants, and the lint rule.

PR #203, opened 2026-05-22
Add app.* repository + audit-routing layer (Invariant 10, PR2)

bsaffel:
## Summary
Every change to a protected `app.*` table (user categories, merchants, rules, budgets, account/balance settings, curation notes/tags/splits, match decisions, …) should write a paired **audit-log row in the same transaction** — so no user state mutates without an audit trail. This PR makes that a *structural* rule rather than a per-call convention: such writes route through a repository (`*Repo`) class that emits the audit row, instead of raw `INSERT`/`UPDATE`/`DELETE`.

This is the first implementation PR of that layer — it sets the `*Repo` pattern every protected table will adopt and migrates the category-taxonomy writers onto it. (The rule is formally "Invariant 10" in `architecture-shared-primitives.md`; full design in `docs/specs/app-integrity-invariant.md`.)

## Changes
- **`BaseRepo`** (`src/moneybin/repositories/base.py`) — shared transaction helper, a single audit-emission point (`_emit_audit`, the only place `app_mutation_audit_emitted_total` increments), full-row `before`/`after` serialization (datetime/Decimal-safe — the captured row is the foundation for future undo), and a `_fetch_one` PK read.
- **`UserCategoriesRepo` + `CategoryOverridesRepo`** own all `app.user_categories` / `app.category_overrides` mutation SQL. `CategorizationService` (via its applier) composes them, so `categories_create` / `categories_set` / `categories_delete` now emit audit rows (they previously bypassed audit). Surfaces thread an explicit `actor` (`cli`/`mcp`).
- **Folded the existing `GSheetConnectionsRepo` onto `BaseRepo`** so the new metric covers every repository and there is one audit-emission pattern (also fixes stale "Invariant 9" → "Invariant 10" references from before the #200 renumber).
- **Doctor invariants** — a reusable `app.*` audit-coverage check (every recent protected row has a matching audit row; sampled by recency + cap, `moneybin system doctor --full` scans all) and an `app.user_categories` `(category, subcategory)` uniqueness check; new `DoctorSettings` config block.
- **Metric** `app_mutation_audit_emitted_total{repository, action}` — a repo that mutates without going through `_emit_audit` shows up as a gap vs `audit_events_emitted_total`.
- **Lint-rule spike outcome** recorded in the spec: ruff has no custom-rule plugin API, so the final PR ships a pytest AST walker covering both inline-literal and assignment-then-execute SQL shapes.

Scope: this is one of a multi-PR sequence. The remaining protected tables (merchants, rules, transaction_categories, settings family, edge writers, curation) get repos in later PRs; the lint rule that enforces "no raw mutation outside a repo" lands last, once every table has coverage. The 6 cascade deletes in `delete_category` to other protected tables stay raw until their repos land.

## Test plan
- `uv run pytest tests/moneybin/test_repositories/` — BaseRepo + both new repos + gsheet regression (audit pairing, full-row before/after, atomic rollback, metric increment).
- `uv run pytest tests/moneybin/test_services/test_categorization_audit.py` — category create/toggle/delete now emit audit.
- `uv run pytest tests/moneybin/test_services/test_doctor_app_integrity.py` — coverage flags bypass rows, passes for repo-mutated rows, respects lookback + `--full`; uniqueness check.
- `uv run pytest tests/moneybin/test_services/test_categorization_service_writes.py` — existing writer behavior unchanged.
- `make check test` green (format, lint, pyright, full unit suite — confirmed locally).


bsaffel:
Thanks for the review — addressed in `6ab7b261`:

**🔴 MUST FIX — audit-coverage JOIN omits `target_schema` (`doctor_service.py`)** + **codex P1 (coverage passes on a prior audit after a later un-audited mutation):** rewrote the check as a `NOT EXISTS` that matches `(target_schema, target_table, target_id)` **and** requires an audit row with `occurred_at >= row.updated_at`. Repo writes set the row's `updated_at` and the audit's `occurred_at` from `CURRENT_TIMESTAMP` in one transaction (equal in DuckDB), so legitimate mutations still pass; a row audited once then mutated by a raw bypass advances `updated_at` past every existing audit and is now flagged. New test: `test_audit_coverage_flags_audited_then_bypassed`.

**🟡 `UserCategoriesRepo.update_active`/`delete` ghost audit on missing row:** both now raise `ValueError` when the row is absent (matching `GSheetConnectionsRepo.delete`), so no phantom `before=None/after=None` audit lands. New tests: `test_update_active_raises_on_missing_row`, `test_delete_raises_on_missing_row`.

**🟡 gsheet `insert`/`update_*` lack the `in_outer_txn` hatch:** added `in_outer_txn` to `insert`/`update_status`/`update_after_pull`/`update_mapping` for a uniform composable contract across every repo mutation method.

**🔵 NIT — uniqueness label ambiguity for `NULL` vs `""` subcategory:** left as-is. `subcategory` is `NULL` (never `""`) for top-level categories by construction, and the string is diagnostic-only; not worth special-casing.

**🔵 NIT — `create_category` duplicate-check + insert not in one transaction:** left as-is. The check/insert were already separate before this PR (this PR only moved the INSERT into the repo), and MoneyBin is single-writer (exclusive write lock per `database-writer-coordination.md`), so the TOCTOU window can't occur in practice. If multi-writer ever lands, the dim-categories uniqueness contract gets revisited then.

If that clears the review, please approve.

bsaffel:
Two more coherence items from re-review + a `/simplify` pass, in `dc4129d7` + `9cba8ba9`:

- **Duplicated missing-row guard** (`/simplify`): the `fetch before-image, raise if absent` block was copied 6× across the two repos — lifted into `BaseRepo._require`; all call sites route through it.
- **`base.py` `repository` ClassVar has no default → silent runtime failure:** added `BaseRepo.__init_subclass__` so a repo that forgets its `repository` label fails at class-definition (import) time, not on first `_emit_audit`.
- **`gsheet` `_AUDIT_TARGET` module-constant vs class-attr inconsistency:** moved it to a class attribute, matching `UserCategoriesRepo`/`CategoryOverridesRepo`.

Kept `in_outer_txn` on the gsheet `insert`/`update_*` methods (the uniform composable-repo contract requested last round; documented in `BaseRepo._transaction`). All green locally (repo + gsheet/connection suites, format/lint/pyright).

bsaffel:
Round 3 — one genuinely-new finding, addressed; the rest are already-resolved re-posts.

**New, fixed (`d6f19fbe`):** `system_doctor` MCP tool now takes a `full` param and passes it to `run_all`, mirroring `moneybin system doctor --full` — the deep audit-coverage sweep is reachable from both surfaces.

**Already resolved in earlier commits (verified in HEAD `d6f19fbe`)** — these are being re-posted against an earlier snapshot:
- 🔴 audit-coverage JOIN omits `target_schema` → fixed in `6ab7b261` (now a `NOT EXISTS` matching `(target_schema, target_table, target_id)` AND `occurred_at >= updated_at`). `grep "a.target_schema = ?" doctor_service.py` → present.
- 🟡 `update_active`/`delete` ghost audit on missing row → fixed in `6ab7b261`, consolidated to `BaseRepo._require` in `dc4129d7`.
- 🟡 gsheet lacks `in_outer_txn` → added in `6ab7b261`.
- 🟡 `repository` ClassVar no default → `BaseRepo.__init_subclass__` fail-fast added in `9cba8ba9`.
- 🟡 gsheet `_AUDIT_TARGET` module-const vs class-attr → moved to class attr in `9cba8ba9`.

**Declined:** the `actor="cli"` default on gsheet `insert`/`update_mapping` is pre-existing (not changed by this PR) and the gsheet actor API is out of scope for the audit-routing work — leaving it. The 🔵 NITs (subcategory `NULL` vs `""`; create-check TOCTOU under the single-writer model) stand as previously noted.

Could you re-review against `d6f19fbe`? I believe everything actionable is resolved.

bsaffel:
Round 4 — `b33efbd9`:

**Fixed:**
- 🟡 `BaseRepo._transaction` `except Exception` → `except BaseException` with commit moved to an `else` clause, so a `KeyboardInterrupt`/`SystemExit` mid-write rolls back instead of leaving the transaction open (re-raised immediately, never swallowed).
- 🟡 `delete_category(force=True)` cascade path now has an audit test.
- 🔵 `soft_disconnect` gains `in_outer_txn` for parity with the other gsheet mutations.

**Already fixed (stale thread):** `system_doctor` *does* take `full` and passes it to `run_all(verbose=False, full=full)` as of `d6f19fbe` — `grep "def system_doctor" system.py` → `system_doctor(full: bool = False)`. The thread is carrying forward against an older snapshot.

**Declined, with rationale:**
- **codex P1 — "validate coverage against writes that don't bump `updated_at`":** the doctor check is a *sampled runtime heuristic* keyed on `updated_at` (the canonical mutation timestamp every repo + writer sets). Catching a write that bypasses **both** the audit row **and** `updated_at` would require row-content hashing or DB triggers (DuckDB has neither) — out of Phase 1 scope (the spec defers coverage-check sophistication to the hosted tier). The structural defense against raw bypass writes is the **lint rule** landing in the final PR, which prevents them at the source rather than detecting them after the fact.
- 🔵 `applier.py:756` discarding the `AuditEvent` from `_user_categories.delete` is intentional — the cascade-threading (`parent_audit_id`) wiring lands with the referenced tables' repos in a later batch; the inline comment notes this.
- gsheet `actor="cli"` default + the create-check TOCTOU (single-writer model): unchanged, as previously noted.

CI was green (6/7, Claude Review finalizing) before this push. I believe everything actionable is resolved — please re-review against `b33efbd9`.

bsaffel:
Addressed in `4a73bbcd` (fix-review pass):

**🔴 MUST FIX — `pk_col` SQL-identifier interpolation:** now quoted via `sqlglot` (`exp.to_identifier(...).sql("duckdb")`) in both `BaseRepo._fetch_one` (column list + pk) and the doctor coverage query — defense-in-depth per `security.md`, not relying on the `noqa`.

**🟡 CONSIDER:**
- `MatchApplier._transaction` now rolls back on `BaseException` and commits in an `else` clause (matches `BaseRepo._transaction`) — interrupt-safety on the `delete_category` cascade.
- `toggle_category` gains `in_outer_txn`, forwarded to both repo branches and through the service facade.
- `actor` is threaded through `GSheetConnectionsService.connect/disconnect/reconnect`; MCP gsheet tools now pass `actor="mcp"` (was logging `"cli"`), CLI passes `"cli"`.

**🔵 NIT:** `NULL` vs `""` subcategory now render distinctly in the uniqueness detail; added a `TODO(Batch B/C)` at the `delete_category` cascade for `parent_audit_id` threading; commented the `create_category` single-writer reliance.

**codex P1 (design — non-blocking per your verdict):** the audit-coverage heuristic keys on `updated_at` over existing rows, so raw bypasses that skip `updated_at` or `DELETE` rows aren't detectable by a table scan. Documented as an explicit limitation in the method docstring; the **lint rule (final PR)** is the structural guard that prevents raw bypasses at the source. Content-based coverage tracked as an M3E follow-up.

`make check` + full unit suite green locally. Please re-review against `4a73bbcd`.

bsaffel:
Addressed in `c8d9a97f`:

**🔴 MUST FIX:**
- `connection_service.disconnect(purge=True)` txn now catches `BaseException` (interrupt-safety), matching `BaseRepo`/`MatchApplier`. Good catch — I'd fixed the other two guards but missed this manual one.
- `GSheetConnectionsRepo.list_all`/`list_healthy` now quote column identifiers via the now-public `quote_ident` — closes the inconsistency I introduced by quoting `_fetch_one` but not the read methods.

**🟡 CONSIDER:**
- Added audit-coverage doctor checks for `app.category_overrides` and `app.gsheet_connections` (both wrapped in this PR) — `_run_app_integrity` now covers all three repo-wrapped tables.
- Documented the `_emit_audit` metric-before-commit trade-off: both counters increment together inside the txn (matching the pre-existing `audit_events_emitted_total`), so a rollback over-counts both by one but never opens a false *gap between* them; best-effort telemetry.

**Deferred (out of scope):** the `reconnect()` stale-`sheet_name` bug is **pre-existing** (in `update_mapping`, not introduced here) and is a gsheet reconnect-correctness issue unrelated to audit-routing — tracked as a follow-up rather than widening this PR.

`make check` + full unit suite green locally. Please re-review against `c8d9a97f`.

bsaffel:
Addressed in `1e88cba3` — thanks for confirming the 6 resolved threads.

**🟡 CONSIDER (the one gating item) — spec drift:** updated `moneybin-mcp.md` §`system_doctor` (the `full` parameter, behavior, `run_all(verbose=False, full=False)` signature, and the `[--verbose] [--full]` CLI invocation) and noted `--full` on the doctor command in `moneybin-cli.md`. Per the `mcp-server.md` two-spec rule.

**🔵 NIT:** `string_agg(category_id, ',' ORDER BY category_id)` — deterministic `affected_ids` ordering.

**On the 5 tracked-open threads:**
- **`EMqRs` (commit in `else` not rolled back on commit failure):** respectfully pushing back — `try/except/else` with commit in `else` is the canonical Python transaction-context-manager idiom. Calling `rollback()` after a *failed* `commit()` risks erroring on an already-aborted transaction and masking the original commit exception. A commit failure is catastrophic-rare (disk/corruption) for a single-writer local DuckDB; the else-form propagates it cleanly. Happy to revisit if you've seen DuckDB leave a half-open txn after a failed commit specifically.
- **`ELZzc` (gsheet repo `actor="cli"` default):** the actual bug (MCP mutations mis-logging as `cli`) is fixed — `actor` is now threaded through the service from every surface. The repo-level default is now only a fallback for direct repo calls; left as-is to avoid churning all gsheet callers/tests, consistent with how the repo's other defaults work.
- **`ELkfs` / `ELxQt` (updated_at + delete-bypass coverage):** documented as explicit design limitations of the sampled heuristic; the lint rule (final PR) is the structural guard; content-based coverage tracked as an M3E follow-up. Confirming explicit deferral.
- **`ELlbo` (discarded AuditEvent):** `TODO(Batch B/C)` marks the call site; cascade `parent_audit_id` threading lands with the referenced tables' repos.

`make check` + doctor suite green. Please re-review against `1e88cba3`.

bsaffel:
@claude — re-review request for `1e88cba3`. All CI is green (CI + Scenarios + AI Code Review workflows all `success`) and I've verified every actionable finding is resolved in HEAD. Could you confirm whether this is ready for approval, and if so approve / clear the stale `CHANGES_REQUESTED`?

**Resolved & verified in `1e88cba3`:**
- 🔴 `pk_col` / column identifiers quoted via `quote_ident` (sqlglot) in `BaseRepo._fetch_one` + the doctor coverage query, and column-quoting in `GSheetConnectionsRepo.list_all`/`list_healthy`.
- 🟡 `connection_service.disconnect(purge)`, `MatchApplier._transaction`, `BaseRepo._transaction` all roll back on `BaseException`.
- 🟡 `actor` threaded through `GSheetConnectionsService.connect/disconnect/reconnect` → MCP logs `actor="mcp"`, CLI `"cli"`.
- 🟡 doctor audit-coverage now covers all three repo-wrapped tables (`user_categories`, `category_overrides`, `gsheet_connections`).
- 🟡 spec drift: `full` parameter documented in `moneybin-mcp.md` §system_doctor + `moneybin-cli.md`.
- 🔵 `string_agg(... ORDER BY category_id)`; NULL/`""` subcategory distinction; `delete_category` cascade `TODO`.

**Intentionally not changed (rationale):**
- `EMqRs` (commit in `else:`): the `try/except/else` form with commit in `else` is the canonical Python transaction-CM idiom; calling `rollback()` after a *failed* commit risks erroring on an already-aborted txn and masking the commit exception.
- gsheet repo `actor="cli"` default: the actual mis-logging bug is fixed (every surface now threads `actor`); the repo default is a harmless fallback for direct calls.
- Coverage of `updated_at`-skipping / delete bypasses: a sampled runtime heuristic can't detect these without triggers/hashing — the lint rule (final PR) is the structural guard; content-based coverage is tracked as an M3E follow-up.
- Remaining 🔵 NITs (`_serialize_for_audit` passthrough, `__init_subclass__` abstract carve-out, `target_id` VARCHAR compare) are non-blocking and defensible for the current (all-VARCHAR-PK, no-abstract-repo) shape.

If anything above is unconvincing, point me at it; otherwise this should be good to approve.

PR #206, opened 2026-05-22
Add app.* repos + audit routing for categorization tables (Invariant 10)

bsaffel:
## Summary

Wraps the categorization-domain `app.*` writers in `*Repo` classes so every
mutation emits a paired `app.audit_log` row in the same transaction
(Invariant 10, per `docs/specs/app-integrity-invariant.md` PRs 3–5). Builds on
the `BaseRepo` foundation from #203.

## Repositories added

- `UserMerchantsRepo` (`app.user_merchants`) — `insert`, `append_exemplar`
- `CategorizationRulesRepo` (`app.categorization_rules`) — `insert`, `deactivate`
- `ProposedRulesRepo` (`app.proposed_rules`) — `insert` / `reinforce` /
  `supersede` / `mark_approved` / `mark_rejected`
- `TransactionCategoriesRepo` (`app.transaction_categories`) — `set` /
  `upsert_guarded` (precedence-guarded) / `clear` / `delete_by_rule`

## Migrations

- The applier merchant/rule/transaction-category writers and the
  `auto_rule_service` proposal/promotion/deactivation sites route through repos.
- Rule promotion threads its audit id onto the paired proposal-approve UPDATE
  (cascade), proven by an `AuditService.chain_for` test.
- `actor` threaded from CLI (`"cli"`) and MCP (`"mcp"`); batch-engine writes use
  `"system"`.

## Doctor

- Audit-coverage for every wrapped table (helper generalized with an
  `updated_col` param for tables that lack `updated_at` —
  `proposed_rules`→`proposed_at`, `transaction_categories`→`categorized_at`).
- `user_merchants` orphan warning; `proposed_rules.rule_id → categorization_rules`
  FK; `transaction_categories.transaction_id → core.fct_transactions` FK.

## Spec reconciliations (documented in the spec)

- **No `RuleDeactivationsRepo`** — `app.rule_deactivations` was dropped in V018.
- **No priority-uniqueness check** — priority is intentionally non-unique
  (deterministic `created_at` tie-break).
- **FK direction corrected** — `proposed_rules.rule_id → categorization_rules`.
- **Audit action unified** on `categorization_rule.deactivate` (taxonomy);
  `category.set`/`.clear` now capture the full row (Req 4), superseding the
  prior column-subset payload.
- **`delete_category`'s 6-table cascade stays raw** until the remaining two
  tables get repos, so it threads `parent_audit_id` atomically in one pass; the
  lint rule lands in the final spec PR.

## Test plan

[...]

PR #208, opened 2026-05-22
Add settings-family app.* repos + audit routing (Invariant 10, PR 6/9/10)

bsaffel:
## Summary

Batch C of the `app.*` integrity work: wraps the settings-family writers — `app.account_settings`, `app.balance_assertions`, `app.budgets` — in audited `*Repo` classes so every protected mutation pairs with an `app.audit_log` row in the same DuckDB transaction. This closes three of the bypass sites where user-state writes previously left no forensic trail. Implements spec PRs 6, 9, and 10 of `docs/specs/app-integrity-invariant.md` (formally Invariant 10).

## Background

`app.*` holds the only non-reconstructible state in MoneyBin (account settings, balance anchors, budgets, categories, rules, …). Audit routing was enforced by convention, not structure, so a bad service mutation could silently corrupt user state with no recovery path. The chosen design is a thin repository layer (rejected alternatives: a `@audited` decorator — still per-method discipline, invisible at SQL-grep time; event sourcing — wrong cost profile for a single-user local app). Each protected table gets a tiny `*Repo` whose mutation methods emit audit unconditionally in the write's transaction, and services compose repos instead of running raw SQL.

This is the third increment of a phased rollout — AI-PR2 (#203) established `BaseRepo` + the first repos; Batch B (#206) covered the categorization family. The settings family (`account_settings`, `balance_assertions`, `budgets`) were bypass sites identified during spec drafting. The undo *consumer* is deliberately deferred to Phase 2; this PR only lays the data foundation (full pre/post-row capture, threadable `parent_audit_id`). The lint rule that structurally forbids raw `app.*` mutations lands last (AI-PR13), after every table has repo coverage.

## Changes

- **Repositories** (`src/moneybin/repositories/`): new `AccountSettingsRepo`, `BalanceAssertionsRepo`, `BudgetsRepo`, each subclassing `BaseRepo` — full pre/post-row capture for `before`/`after`, audit emitted via the single `_emit_audit` point. Audit actions follow the `<entity>.<verb>` taxonomy: `account_settings.set`/`.delete`, `balance_assertion.set`/`.delete`, `budget.set` (insert + update branches).
- **Services**: `account_service`, `balance_service`, `budget_service` compose the repos instead of executing raw `INSERT`/`UPDATE`/`DELETE`. The pre-existing `AccountSettingsRepository` data-access class (a non-audited near-namesake) was removed — its read folded into `AccountService._load_settings`, leaving one repository per table.
- **Doctor** (`doctor_service.py`): per-table audit-coverage + foreign-key checks (`account_settings`/`balance_assertions` → `core.dim_accounts`; `budgets` → `core.dim_categories`, NULL-skipping). Total invariants 16 → 22. `balance_assertions`' composite `(account_id, assertion_date)` PK maps to a composite audit `target_id`; the coverage helper gained a `pk_expr` hook to project the matching key.
- **Reviewer focus**: the composite-PK ↔ `pk_expr` coupling in `balance_assertions` (repo `_target_id` and doctor `_BALANCE_ASSERTIONS_PK_EXPR` must agree); the two deferred repo imports in `account_service`/`balance_service` `__init__` (break a real circular import — these modules are eagerly re-exported by `services/__init__`, so a repo-first import order otherwise hits a partially-initialized module). A dated spec reconciliation note documents the five spec↔code drifts resolved.

## Test plan

- `make check test` green: format, lint, full pyright (incl. tests), 3159 unit tests; e2e (`-k "balance or budget or account or settings"`) and touched scenarios pass.
- New per-repo unit tests (`tests/moneybin/test_repositories/test_{account_settings,balance_assertions,budgets}_repo.py`): insert/update/delete happy paths, full-prior-row capture, `parent_audit_id` threading, metric increment, rollback-when-audit-raises.
- Doctor integration tests (`test_doctor_app_integrity.py`): coverage pass/flag — including the composite-key bypass path — and FK flag/pass for all three tables; invariant-count test bumped to 22.

## Breaking changes

- Internal API only (no user-facing surface change — no new/renamed CLI commands or MCP tools): the migrated service methods now require an `actor` argument — `AccountService.settings_update`, `BalanceService.assert_balance` / `delete_assertion`, `BudgetService.set_budget`. All in-tree callers (CLI, MCP) are updated in this PR; the deprecated `AccountService` delegates (`rename`/`archive`/`unarchive`/`set_include_in_net_worth`) default `actor="system"`. Mutations of these three tables now emit an `app.audit_log` row that previously did not exist (additive forensic data).

## Related

- Follows #203 (BaseRepo + first repos), #206 (Batch B: categorization family)
- Spec: `docs/specs/app-integrity-invariant.md` (PRs 6, 9, 10)


bsaffel:
@claude Thanks — addressed in `946567d1`:

- **🔴 `budget_service.py` `start_month`** — fixed. The overlap query now selects `start_month` too, and the `updated` path returns the **stored** window (`existing[1]`), not the request argument. Added `test_update_cross_month_overlap_reports_stored_start_month` (stored `2026-01`, request `2026-06` → result + DB row both report `2026-01`).
- **🟡 `delete_assertion` validation** — resolved via your second option (document the contract). Deleting is deliberately forgiving/idempotent: an unknown `account_id` is a silent no-op, asymmetric with `assert_balance` by design (you can't *create* an anchor for a nonexistent account, but *removing* one is best-effort). This contract is locked by `test_e2e_mutating.py::TestBalanceAssertions::test_balance_delete_nonexistent_is_noop` (deletes `nonexistent_acct`, expects exit 0) — adding the check would have broken it. Docstring now states this explicitly; added a unit test for the no-op.
- **🟡 CLI delete test** — strengthened: `assertion-delete` now inspects `call_args` for `account_id` and `actor="cli"`.
- **🔵 `budgets_repo.py` `CURRENT_TIMESTAMP`** — added the requested comment marking it intentional in a plain UPDATE (no `NOW()` cargo-culting).
- **🔵 deferred-import comment** — respectfully, the cycle **is** real. It doesn't route through `audit_service.py`'s own imports; it routes through *package init*: importing `moneybin.services.audit_service` runs `services/__init__.py`, which eagerly imports `account_service`, which imports the repo. On a repo-first import order (any repo test module) a module-level import here fails with `ImportError: partially initialized module` — verified empirically (module-level → ImportError; deferred → clean). `budget_service` can import its repo at module level only because it is *not* re-exported by `services/__init__`. The committed comment (this PR also folded in a clarified version of it) now spells out that mechanism.

CI was green before this push; re-running now. If this clears your review, please approve.

bsaffel:
@claude Round 2 (against `946567d1`) — the round-1 🔴/🟡 are resolved (that earlier review re-ran against the stale `c82875a`; both were already fixed in HEAD). Two new inline findings on `946567d1`, addressed in `2320b2cb`:

- **🟡 `settings_update` phantom audit on empty diff** — good catch, fixed. Added an early-exit `if not diff: return current, warnings` before the repo call, so a no-op (all-`None`) call no longer bumps `updated_at` or emits an `account_settings.set` audit row. Added `test_empty_diff_is_noop_no_audit`.
- **🔵 `budget_id` generated outside the `with self._transaction` block** — declining, intentionally. This matches the established sibling pattern: `UserMerchantsRepo.insert` and `CategorizationRulesRepo.insert` both generate `uuid.uuid4().hex[:12]` before opening the transaction. Moving it inside would introduce a parallel pattern for one repo (coherence regression). There's no real leak — the uuid isn't persisted until the INSERT, so a PK collision (~0.002% at 10k) just rolls back and discards an in-memory string. Keeping it consistent with the other repos.

CI green on `946567d1`; re-running on `2320b2cb`. If this clears your review, please approve.

bsaffel:
@claude Round 3 addressed in `a9501583`:

**Fixed:**
- **🔴 rollback tests** — all three now use `with pytest.raises(RuntimeError):` instead of bare `try/except`, so a regression that stops raising fails loudly.
- **🟡 `delete_assertion` no-op log** — now branches on the repo's return value; logs "No balance assertion to delete…" on the no-op path.
- **🟡 settings_update no-op vs unknown account** — verified `_assert_account_exists` runs *before* the `if not diff` early-return (it does); added `test_empty_diff_still_rejects_unknown_account` to lock it.

**Declined, with reasons:**
- **🟡 deferred-import "fix root cause"** — the `services/__init__` re-export of `AccountService`/`BalanceService` is intentional public API. The deferred import is the correct solution, not a workaround; there is no root cause to remove.
- **🟡 `_BALANCE_ASSERTIONS_PK_EXPR` cross-ref** — already present: the constant's comment reads "Must match `BalanceAssertionsRepo`'s `target_id`".
- **🟡 `_require` → `ValueError` vs `UserError`** — unreachable via `set_budget` (the `budget_id` is read in the same call immediately before `update`), and `_require` is the shared `BaseRepo` guard used by every repo; special-casing it here would fork the pattern.
- **🔵 pyright-comment placement** — follows the prescribed "after docstring, before `from __future__`" convention; **🔵 positional/keyword "asymmetry"** — `set` takes `account_id`/`assertion_date` positionally too, so there's no asymmetry.

Two items are deliberately **out of scope** for this audit-routing batch and tracked as follow-ups: the pre-existing `budget_service.status()` category fan-out (a read-path correctness issue with no DB uniqueness constraint — orthogonal to Invariant 10) and the cosmetic `_run_app_audit_coverage` docstring wording.

CI green across all functional checks. Note: these last two rounds surfaced new findings on unchanged code each pass (the rollback tests have been present since the first commit) — further rounds will keep doing so. The substantive findings are resolved.

bsaffel:
@claude Cosmetic NITs addressed in `042c8485`:

- **`_run_app_audit_coverage` docstring** — corrected: the check name derives from `table_ref.name`; `pk_col` is only the single-column projection fallback used when `pk_expr` is `None`.
- **`set_budget` positional access** — the existence-check row is now unpacked into named locals (`existing_budget_id`, `existing_start_month`) so a future SELECT-list change can't silently shift positional access.

All earlier 🔴/🟡 remain resolved (verified against HEAD each round). The one remaining open item — `budget_service.status()` category fan-out — is **pre-existing read-path code** not touched by this audit-routing PR; tracked as a follow-up. CI green.

PR #209, opened 2026-05-22
Add agent-callable transaction match accept/reject

bsaffel:
## Summary
Accepting or rejecting a proposed transaction match (dedup or transfer pair) previously existed only in the interactive CLI review queue, so agents and scripts couldn't reach it. This adds a non-interactive accept/reject surface across MCP and CLI, routed through `MatchingService`, with structured recovery actions on every error path.

## Background
Two design decisions worth flagging for review:
- **Read-tool naming.** The pending-list tool is `transactions_matches_pending`, not the `transactions_matches_review` the contract spec originally documented — renamed for coherence with the existing `transactions_categorize_pending`, so there is one pattern for "fetch the pending queue." The spec was reconciled to match.
- **Write path.** `app.match_decisions` mutations route through `MatchingService` over the existing persistence primitive rather than a dedicated repository wrapper. That table is not yet covered by the app-state mutation-routing convention; wrapping it is left to that follow-up work, so this PR adds no new raw write.

## Changes
- **Service** (`MatchingService`): `set_status` accepts/rejects one decision by `match_id`, enforcing the pending→accepted/rejected transition and returning structured `recovery_actions` on invalid transitions or unknown ids (rejecting an already-accepted match points the caller at `transactions matches undo`); plus `get_pending`, `accept_all_pending`, and a `get_match_decision` read helper.
- **MCP**: four tools complete the matches surface — `transactions_matches_set`, `transactions_matches_pending`, `transactions_matches_run`, `transactions_matches_history`.
- **CLI**: `moneybin transactions matches set <id> --status accepted|rejected`, plus the now-wired `transactions review --type matches --confirm/--reject/--confirm-all` (previously stubbed).
- **Docs**: matches read tool documented under the established `_pending` name (was `_review`); MCP/CLI/capabilities specs + CHANGELOG updated.

## Test plan
- `make check test` green (3151 passed, 3 expected skips) after merging current `main`.
- Service unit tests: accept/reject transitions, idempotent re-assertion, reject-already-accepted error, unknown-id error.
- CLI e2e: `matches set` (accept, reject, invalid-status usage error), `review --type matches --reject`, `--confirm-all`.
- MCP e2e: registration + invocation of set / pending / run / history.


bsaffel:
Addressed the latest review in `e27ebfd2`. Two notes on judgment calls:

**Keeping `idempotent=True` on `transactions_matches_set` (deliberate).** The MCP `idempotentHint` contract is "safe to retry without side effects" — it concerns *repeating the same call*, not whether every possible transition succeeds. `set(X, "accepted")` repeated is a no-op → idempotent. `set(X, "rejected")` after an accept is a *different* call whose failure is a state-transition constraint, orthogonal to idempotency. This is also the surface-design contract for `_set` (shape 1a/1b are idempotent by definition). I sharpened the description + spec to make the no-op-vs-error distinction explicit rather than flip the annotation. (`transactions_matches_run`, by contrast, is now correctly `idempotent=False` — re-running genuinely proposes a different pending set.)

**Two CONSIDER items deferred with tracked follow-ups:**
- `system_audit_undo` recovery target — intentionally retained; it's the near-term M2D undo tool, and the `rationale` already names the CLI fallback for the interim.
- Non-atomic `--confirm X --reject Y` (different IDs) — the same-ID contradiction is now guarded; cross-ID atomicity needs a transaction-aware service variant (DuckDB has no nested transactions), tracked as follow-up work rather than expanded in this PR.

NITs (docstring dead-end, conftest `try/finally`, `matches_run` annotation) all fixed in the same commit.

PR #210, opened 2026-05-22
Add AI consent ledger (CLI + MCP grant/revoke/status/log)

bsaffel:
## Summary
MoneyBin's AI consent ledger: a persisted, user-controllable record of which
AI feature categories the user has authorized for which backend, surfaced on
both CLI and MCP, with paired audit-log rows and `privacy.log` events. This is
PR 3 of the privacy data-classification rollout (PR 2 shipped the
redaction/introspection middleware). It ships the **ledger only** — the
data-withholding enforcement gate (degrading responses without consent) is
deliberately deferred until a consumer needs it (hosted multi-user, or a
direct cloud-AI feature). Always-on CRITICAL masking shipped in PR 2 and is
unchanged.

## Background
The privacy spec describes an enforcement gate that degrades medium/high
responses to aggregates until consent is granted. That gate is deferred (decided
during build): for local single-user MCP the AI host's own tool-approval dialog
already gates each call, so a second degrade-to-aggregate gate is mostly
friction; and the gate's real beneficiaries (hosted/multi-user, direct cloud-AI
features) don't exist yet. The ledger delivers everything else — the consent
receipt, user controls, audit trail, and privacy metadata on every response.
`redact_typed`, the `@mcp_tool` decorator, and all response shapes are unchanged.

## Changes
- **Schema:** new `app.ai_consent_grants` table (schema DDL for fresh installs +
  V022 migration for existing DBs), classified in the privacy registry;
  `grant_prompt` is `DESCRIPTION`-tier and omitted from read payloads so they
  stay LOW.
- **Data layer:** `ConsentRepo` pairs every mutation with an `app.audit_log` row
  (Invariant 10); `ConsentService` resolves the backend, validates the feature
  category, and emits fail-soft `privacy.log` events only on real state changes
  (an idempotent re-grant logs nothing and reports `noop`).
- **Surfaces:** `moneybin privacy grant/revoke/revoke-all/status/log` CLI +
  `privacy_consent_grant`/`privacy_consent_revoke`/`privacy_status`/`privacy_log`
  MCP tools, both over the same service with `--output json` parity and LOW-tier
  typed payloads.
- **Config:** minimal `AIConfig` (`default_backend`, `consent_policy` in
  standard/strict). `consent_policy` (backend re-prompt policy) is named
  distinctly from each grant's `consent_mode` (persistent/one-time).
- **Folded-in PR-2 followups + docs:** redundant balance `sensitivity` removed,
  frozenset-redaction regression test, privacy/MCP/CLI/capabilities specs +
  CHANGELOG updated.

## Test plan
- [ ] `make test` green (3179 passed locally after merging `main`).
- [ ] `make check` (ruff + pyright) clean.
- [ ] V022 creates `app.ai_consent_grants` idempotently; CHECK rejects an invalid `consent_mode`.
- [ ] Repo: grant/revoke/revoke_all each pair an `app.audit_log` row; idempotent grant emits no second audit row; revoke_all captures before-images pre-UPDATE.
- [ ] CLI: `privacy grant` then `status --output json` reflects it; `revoke` clears it; `log` shows `consent.grant`.
- [ ] MCP: grant/revoke/status/log tools register and round-trip against a temp DB.
- [ ] `privacy_status` returns `consent_policy` (standard/strict) distinct from per-grant `consent_mode`.

## Breaking changes
None requiring action. `app.ai_consent_grants` is additive — V022 auto-applies on
next DB open; fresh installs get it from the schema DDL. New
`MONEYBIN_AI__CONSENT_POLICY` env var defaults to `standard`.

## Related
- Follows #192 (privacy middleware, PR 2).

Two minor PR-2 follow-ups (error-path `cli_actor`, `import_inbox` typed payloads)
remain tracked separately as a future PR.


bsaffel:
Thanks for the reviews — addressed in `730be196`.

**Codex P1 (read-only status before migration) — fixed.** Read-only `Database` opens return before `init_schemas()`/migrations (`database.py:302-313`), so `privacy status` on a freshly-upgraded profile (before any write-mode open) could hit a catalog error. `ConsentRepo.list_active`/`list_all` now short-circuit to empty when `app.ai_consent_grants` is absent. Regression test: `test_list_tolerates_missing_table`.

**Codex P2 (default_backend sentinel) — fixed.** `PrivacyStatusPayload.default_backend` is now `str | None`; an unset backend serializes to JSON `null` instead of `"(none)"`, so it can't be fed back into grant/revoke as a real backend. Text output still renders `(none)` for humans. Test asserts `default_backend is None` on empty status.

**Claude — verb-before-entity naming — fixed.** Renamed `privacy_grant_consent` / `privacy_revoke_consent` → `privacy_consent_grant` / `privacy_consent_revoke` (verb-last `<entity>_<verb>` per `surface-design.md`), updated across the tools, `moneybin-mcp.md` §11, the capabilities map, and CHANGELOG.

**Claude — `write_privacy_event` exception handling — respectfully declined (already handled).** `write_privacy_event` is already fail-soft: its entire body is wrapped in a top-level `except Exception` (`src/moneybin/privacy/log.py:199`) that logs at WARNING and never re-raises (the inner `raise` at :198 re-raises into that same outer handler). The call sites cannot propagate a log-write failure, so a second try/except would be unreachable. Can add an inline comment noting the fail-soft contract at the call site if that reads clearer.

If these resolve your reviews, please approve.

bsaffel:
Round 2 addressed in `324c4db4` (0 🔴 MUST FIX confirmed; these are the 🟡 CONSIDER items).

**Fixed:**
- **Prompt shows resolved backend + one-time caveat** (Codex P2 / Claude #1): the `privacy grant` confirmation now displays the backend that will actually be recorded (resolved default, not the literal `default`) and, for `--mode one-time`, notes that enforcement is pending so the grant persists until revoked. MCP `privacy_consent_grant` docstring carries the same note.
- **Removed unused `grant_prompt` param** (Claude #3) from `ConsentService.grant_consent` — the prompt is always built internally; no caller passed it.
- **`last_n` → `last`** (Claude #4): MCP `privacy_log` param renamed to match the CLI `--last` flag and the spec.
- **`privacy log` text output now includes `backend`** (Claude #5).
- **Added `ConsentService.revoke_all` service test** (Claude #6): count, empty-ledger, and idempotent-second-call.

**Already fixed (stale re-flag):** Codex P1 (read-only `privacy_status` catalog error). The fix lives one layer down — `ConsentRepo.list_active`/`list_all` short-circuit to empty when `app.ai_consent_grants` is absent (`_table_exists` guard, `consent_repo.py`), with `test_list_tolerates_missing_table` covering it. The bot re-anchored the comment to the `get_database(read_only=True)` line at `status.py:18` without tracing into the repo guard, so `privacy_status` returns an empty ledger rather than erroring.

**Respectfully declined:** Claude #2 (`revoke_all` logs a single wildcard `privacy.log` event). This is deliberate: `privacy.log` is best-effort flow telemetry, while the authoritative per-grant record lives in `app.audit_log` — `revoke_all` already emits one audit row per revoked grant (with before/after images). A wildcard `revoked-all` telemetry summary plus full per-grant audit detail is the intended split; emitting N telemetry events for one user action would add noise without adding the authoritative detail (which is already in the audit log).

If these resolve the reviews, please approve.

bsaffel:
Pushed `7ffb98a6` (review fixes + clean merge of `origin/main`). Round-3 🟡 triage, mapped to the numbered findings:

**1. ✅ `PrivacyLogRow` dropped `tool_call` fields** — added `sensitivity` / `row_count` / `classes_returned` (LOW-tier, defaulted) plus a `PrivacyLogRow.from_event` constructor. CLI text and MCP now render `tool_call` events meaningfully instead of emitting empty `{feature_category:"", backend:""}` rows.

**2. ✅ `_emit_audit` before the not-found guard in `grant()`** — the `after is None` guard now precedes `_emit_audit`, so the impossible path can't record an `after=null` audit row or bump the audit metric before the rollback.

**3. ⏸️ Deferred — falsy `if backend:` / `if default:`** — keeping the falsy check rather than `is not None`. An empty string is not a valid backend identifier: `if backend:` correctly treats `""` as unspecified and falls through to the default (or errors when none is configured). `is not None` would *regress* — `backend=""` would create a grant with a blank recipient, and an empty `MONEYBIN_AI__DEFAULT_BACKEND` would be accepted as a real backend. The current behavior is intended; documented inline on `resolve_backend`.

**4. ✅ Confirmation prompt shown before backend could resolve** — `privacy grant` and `privacy revoke` now resolve + validate (via new `ConsentService.resolve_backend` / `validate_category` statics) *before* prompting. The user confirms the actual backend, and an unconfigured backend errors before any consent is requested. This also covers **Codex P2** (`revoke.py`) — the revoke prompt now names the resolved backend.

**5. ✅ `privacy_log(last=N)` uncapped** — `read_privacy_events` caps at 1000 (one enforcement point for both the CLI `--last` flag and the MCP `last` param); noted in the tool docstring.

**6. ✅ `revoke_all` single wildcard event** — now emits one `privacy.log` event per revoked grant, so a bulk revoke is reconstructable from `privacy log` alone (matches single-revoke and the per-grant audit rows the repo already writes).

**Also addressed:**
- Round-1 migration nit — `COMMENT ON COLUMN` identifier is now `sqlglot`-quoted (security.md defense-in-depth).
- NITs — `ConsentMutationPayload.action` is now a `Literal["granted","revoked","noop"]`; `revoke-all` prints a neutral message at count 0; added service tests for per-grant revoke events and one-time-mode persistence; hoisted inline test imports.

**NITs deferred (not defects):**
- *Stale `privacy_audit` / `privacy_audit_list` doc refs* — these describe a planned, **distinct** forensic tool over `app.audit_log` filtered by `action='ai.external_call'`, not a stale pointer to `privacy_log` (which reads the JSONL event stream). Renaming would conflate two surfaces; left as forward-looking design.
- *`privacy log` uses `typer.echo` vs `grant`/`revoke` `logger.info`* — intentional role split: `status`/`log` render data rows to stdout via `typer.echo` (per cli.md), mutations emit status via `logger.info`.

CI is green on the merged tree (`make check` + 3211 passed, 3 skipped). @claude @codex — if this resolves your findings, please approve.

bsaffel:
Pushed `a41380de`. Triage of the latest round (Claude review on `324c4db4` — which predated the `7ffb98a6` fix push — plus Codex inline comments on the fix commits):

**🔴 Fixed**
- **`test_cli_log_after_grant` not isolated to `tmp_path`** — the privacy CLI `db` fixture now patches `_resolve_privacy_log_dir` to `tmp_path/privacy_log`, matching `test_log.py` / `test_cli_output_redaction.py` / `test_decorator_integration.py`. All grant/revoke/log tests are now xdist-safe and leave no shared state.

**🟡 Fixed**
- **Falsy `if backend:`** — reconsidered the round-3 deferral. `resolve_backend` now **rejects** an empty/whitespace backend as invalid input (raises `MUTATION_INVALID_INPUT`) rather than silently using the default. This addresses the silent-fallthrough concern *without* the `is not None` regression (which would record a grant against a blank `""` recipient). `None` still means "use the default." Test added.
- **`PrivacyLogRow` `""` vs `null`** — cross-type fields are now `str | None` / `int | None`, defaulting to `None` (honest "not applicable to this event shape"). `action` remains the discriminator consumers branch on.
- **`revoke_all` snapshot (Codex P2)** — `ConsentRepo.revoke_all` now returns the revoked `GrantInfo` set captured inside the transaction; the service emits per-grant log events from that authoritative list instead of a second `list_active()` snapshot.

**🟡 Deferred (with rationale)**
- **`revoke` error ordering when no backend + no grant** — `revoke` targets a specific `(category, backend)` grant; when the backend can't be resolved the command is underspecified, so "specify a backend" (with a hint to pass `--backend` or set a default) is the correct, actionable error. The common case (a default backend configured) never hits this. Special-casing "nothing to revoke" would reorder resolution against an existence check across the CLI/service boundary — disproportionate for that path.
- **Atomic uniqueness on active `(feature_category, backend)` (Codex P1)** — MoneyBin serializes writes via a per-database exclusive lock (short-lived write-mode connections; see `database-writer-coordination.md`), so the check-then-insert in `grant()` is race-free — two grant calls can't hold the write lock concurrently. This is the documented design in the `ConsentRepo` docstring; a DuckDB partial index isn't used because support is version-dependent.

**NIT fixed**
- `privacy-and-ai-trust.md` now marks the `privacy_audit` surface as planned/not-yet-registered, notes it's distinct from the shipped `privacy_log`, and fixes the stale `§1089` → `§11` anchor.

**Verified non-issues (re-anchored zombies, no change)**
- *`write_privacy_event` not fail-soft (🔴 inline)* — `write_privacy_event` wraps all I/O in a top-level `try/except` (`log.py:199`) and never raises; the service calling it without an extra `try` is correct. This was conceded in the round-2/3/4 formal verdicts (none list it as MUST FIX); the inline comment is re-anchored from round 1.
- *`privacy_status` read-only schema bootstrap (Codex P1)* — handled by `ConsentRepo._table_exists()`, which short-circuits `list_active()`/`list_all()` to `[]` when the table is absent (read-only opens skip migrations). Covered by `test_list_tolerates_missing_table`.

CI green on the merged tree (`make check` + full suite). @claude @codex — if this resolves your findings, please approve.

bsaffel:
Pushed `7c219856`. Triage of the latest review (0 🔴 + 3 🟡 on `a41380de`):

**🟡 Fixed**
- **`consent_mode` dropped from `privacy_log`** — `PrivacyLogRow` now carries `consent_mode`; `from_event` extracts it and the CLI text renderer shows it. An auditor can now see persistent vs one-time for historical grants. Test asserts the round-trip.
- **Invisible truncation** — `privacy_log` now sets `has_more` (via a `total_count` floor) when a read hits the 1000-row cap, so a capped result is distinguishable from a short log. CLI `--last` help notes the cap.
- **NIT `if actor:`** — both surfaces now use `if actor is not None:` so an explicit empty-string filter isn't silently treated as "return all".

**🟡 Under decision (not a parity violation)**
- **No `privacy_consent_revoke_all` MCP tool** — functional parity holds: bulk revoke is reachable on MCP today via `privacy_status` + per-grant `privacy_consent_revoke`. A dedicated 1-call bulk tool is an ergonomic improvement, not a missing capability. I'm checking with the maintainer whether to add it in this PR or as a fast-follow; will update here.

**Standing deferrals (rationale unchanged, accepted-deferral per your contract)**
- *`revoke` error ordering (no backend + no grant)* — `revoke` targets a specific `(category, backend)`; when the backend can't be resolved the command is underspecified, so "specify a backend" is the correct, actionable error.
- *Atomic uniqueness on active `(feature_category, backend)`* — writes are serialized by a per-database exclusive lock (short-lived write-mode connections), so the check-then-insert in `grant()` is race-free; documented in the `ConsentRepo` docstring.
- *`privacy_audit` references* — a planned, distinct forensic tool (now marked not-yet-registered, anchor fixed), not the shipped `privacy_log`.

CI green on the merged tree last round; full suite passes locally. @claude @codex — with the above, the only open item is the `privacy_consent_revoke_all` scope decision.

bsaffel:
Decision on the one open item: **`privacy_consent_revoke_all` is deferred to a fast-follow PR** (tracked). Functional parity holds today — bulk revoke is reachable via `privacy_status` + per-grant `privacy_consent_revoke` — so this is a 1-call ergonomic upgrade, not a missing capability, and doesn't gate the consent ledger.

That resolves every round-5 🟡 (fix or accepted deferral) with 0 🔴. @claude @codex — please re-review; if the deferrals are acceptable, please approve.

bsaffel:
Pushed `24913699`. Round-6 triage (0 🔴 + 5 🟡):

**🟡 Fixed (2 new)**
- **`resolve_backend` whitespace** — now strips the backend before storing/returning, so a padded `" anthropic"` grant matches a canonical `"anthropic"` revoke instead of becoming unrevocable. Test added.
- **`consent_policy` default** — added a code comment: the field is inert until the enforcement gate ships; the enforcement PR must make the secure-vs-ergonomic posture an explicit decision rather than silently inheriting `"standard"`. (Kept `"standard"` for now since the value does nothing until enforcement exists; flipping pre-empts the enforcement design.)

**🟡 Resolved threads (3 — were stale)**
- **`consent_mode` in `privacy_log`** and **`has_more` truncation signal** were both fixed in `7c219856` (verified in HEAD: `PrivacyLogRow.consent_mode` + `from_event`; `total_count` floor when the cap is hit). The threads were unresolved-but-addressed; I've resolved them.
- **`privacy_consent_revoke_all`** — deferred to a fast-follow by maintainer decision (functional parity holds via `privacy_status` + per-grant `privacy_consent_revoke`). Thread resolved as accepted deferral.

**🔵 NIT (non-blocking, acknowledged)**
- `privacy_log last` has no `ge=1` — `read_privacy_events` already returns `[]` for `max_rows <= 0`, so a non-positive `last` yields an empty payload (no crash). Can add `Field(ge=1)` in the fast-follow.
- V022 has no DB-level `UNIQUE` on active grants — intentional (DuckDB partial-index version-dependency; documented in the module docstring). Writer serialization makes the Python-side check race-free; `sql_query` is operator territory.

That clears every 🔴 and 🟡 (fix, verified-already-fixed, or accepted deferral). @claude @codex — please re-review; if the deferrals are acceptable, please approve.

bsaffel:
Pushed `553b05b8`. Round-7 triage (2 🔴 + 6 🟡):

**🔴 Fixed**
- **No E2E subprocess tests for the privacy CLI commands** — added them: a `grant → status → log → revoke` cycle and a `revoke-all` test in `test_e2e_mutating.py` (real subprocess + encrypted DB), plus `privacy status` / `privacy log` read-path tests in `test_e2e_readonly.py`. All five commands now have subprocess E2E coverage (4 new E2E tests, all passing).
- **No grant → revoke → re-grant cycle test** — added `test_grant_revoke_regrant_cycle`: asserts the re-grant mints a distinct `grant_id`, leaves exactly one active row, and produces two `consent.grant` + one `consent.revoke` audit rows.

**🟡 Fixed**
- **Verb-exception not documented** — both MCP descriptions now state why `_grant`/`_revoke` is used instead of `_set`/`_delete` (authorization event / withdrawal semantics), per surface-design.md's defended-exception rule.
- **No MCP error-envelope test for missing backend** — added `test_grant_no_backend_returns_error_envelope`: asserts `env.error.code == MUTATION_INVALID_INPUT` (a well-formed envelope, not an exception).
- **`has_more` boundary** — `capped` now triggers at `last >= MAX_LOG_ROWS` (was `>`), so the exact-cap case (`last == 1000`) is flagged.

**🟡 Resolved threads (were stale — already fixed in `24913699`)**
- **`resolve_backend` default whitespace** (`PRRT…U5z`) — HEAD already strips the configured default: `if default and default.strip(): return default.strip()`. Resolved.
- **`consent_policy` default undocumented** (`PRRT…U_X`) — HEAD has the rationale comment (inert until enforcement; enforcement PR must choose the posture). Resolved.
- **`privacy_consent_revoke_all`** (`PRRT…Tj2`) — deferred to a fast-follow by maintainer decision (functional parity holds). Resolved as accepted deferral.

That clears both 🔴 and all 🟡. @claude @codex — please re-review.

bsaffel:
Pushed `db6fa6d4`. Round-8 triage (0 🔴 + 3 🟡 — CI was fully green on the prior commit):

**🟡 Fixed**
- **`revoke()` / `revoke_all()` missing after-image None-guard** — both now guard the `_fetch_row` result and raise on the impossible `None`, matching `grant()`. A revocation can no longer write an `after_value=null` audit row that would read as a deletion. (Coherence with the `grant()` guard.)
- **`has_more` only fired at the server hard-cap** — `privacy_log` now signals `has_more` whenever a full page is returned, so `last=50` with 60+ events reports `has_more=True` (not just the `last >= 1000` case). Test added (`test_log_signals_truncation_on_full_page`).
- **Idempotent re-grant silently drops a mode change** — documented in the repo `grant()` docstring and the MCP tool description: re-granting an active category+backend is a no-op and does NOT change `consent_mode`; revoke then re-grant to change it. (consent_mode is inert until the deferred enforcement gate, so this is a documentation fix rather than new mode-change logic.)

**🔵 NIT**
- `last` lower-bound: now safe by construction — the tool guards `effective > 0`, and `read_privacy_events` short-circuits `max_rows <= 0` to `[]`, so `last=0`/negative yields an empty page with `has_more=False`.
- PR-description typo (`privacy_grant_consent` → `privacy_consent_grant`): fixing the PR body; code/spec were already correct.

All four open inline threads resolved. CI was green across all 7 checks (Unit, E2E, Integration, Scenarios) on the prior commit; re-running on `db6fa6d4`. @claude @codex — please re-review.

PR #212, opened 2026-05-23
Add app.* repos + audit routing for edge writers (Invariant 10)

bsaffel:
## Summary

Wraps the three protected `app.*` writers that live outside `services/` —
`app.tabular_formats`, `app.match_decisions`, and `app.imports` — in `*Repo`
classes so every mutation emits a paired `app.audit_log` row in the same DuckDB
transaction (Invariant 10). These were the last edge-writer bypass sites; the
categorization, settings, and balance/budget families already route through repos.

## Background

`app.*` holds the only non-reconstructible state in MoneyBin, and audit routing
was enforced by convention rather than structure. Two reconciliations against the
spec surfaced during implementation:

- The `app.imports` labels write was *already* audited (per-label
  `import_label.add`/`import_label.remove` events), not the silent bypass the spec
  assumed. `ImportsRepo` now emits a single full-row `import.set` audit instead,
  matching every other repo and capturing the complete before/after row.
- `app.match_decisions` stores source-native transaction ids, which don't resolve
  in `core.fct_transactions`; the applicable FK check is
  `account_id`/`account_id_b → core.dim_accounts`.

## Changes

- **Repositories:** `TabularFormatsRepo` (`set`/`delete`), `MatchDecisionsRepo`
  (`insert`/`update_status`/`reverse`), `ImportsRepo` (`set`). All subclass
  `BaseRepo`; full before/after capture; rollback-on-audit-failure.
- **Writer migration:** `extractors/tabular/formats.py` delegates to the repo;
  `matching/persistence.py` keeps reads only (the matcher + `MatchingService`
  compose the repo); `ImportService` label methods use `ImportsRepo`. `actor`
  threaded from CLI/MCP, defaulting to `"system"` for the automated matcher and
  auto-saved formats.
- **Doctor:** audit-coverage checks for the three tables + a
  `match_decisions -> core.dim_accounts` FK check (invariant count -> 26).
- **Audit shape:** `app.imports` audit moves from per-label events to a single
  full-row `import.set` (pre-launch; no migration). Spec, transaction-curation,
  and moneybin-mcp updated.

## Test plan

- [ ] `make check` — ruff format/lint + pyright clean
- [ ] `make test` — full unit suite green (3181 passed, 3 skipped)
- [ ] New repo unit tests cover mutation + paired audit + full before/after + `parent_audit_id` + metric increment + rollback-on-audit-raise
- [ ] Real-DB doctor tests: coverage smokes + `match_decisions` FK flag/pass; invariant count = 26
- [ ] Matcher still writes decisions (`test_dedup_integration`, `test_transfer_integration`); import-label add/remove/set round-trip


bsaffel:

PR #219, opened 2026-05-24
Add audit-log undo consumer (system_audit_undo/history/get)

bsaffel:
## Summary

Ships the audit-log undo consumer: `system_audit_undo`, `system_audit_history`, and `system_audit_get` as MCP tools with full CLI parity (`moneybin system audit undo|history|get`). Any audited `app.*` mutation is now reversible as a unit keyed on `operation_id` — each row's inverse is synthesized from its full audit before/after image and routed back through the `*Repo` layer, so the undo is itself audited (`is_undo`/`undoes_operation_id`) and undoable in turn.

## Background

Invariant 10 already captured a full before/after image for every protected `app.*` mutation. This is the consumer that turns that trail into a reversible history. The shaping decision is **block-don't-cascade**: when a later operation touched the same rows, undo refuses and returns the blocker operations for the agent to walk explicitly, rather than silently reversing unrelated later work — the kind of magic that loses user trust. To make the inverse generic, notes/tags/splits were moved from raw SQL onto dedicated repos first, so every audited table has one.

## Changes

- **Generic inverse (`BaseRepo.undo_event`)** — derives insert→delete, update→restore-before, delete→reinsert from `pk_columns` + the full row image; binds JSON/array columns natively, so no per-repo overrides (incl. `match_decisions`).
- **Dispatch + service** — a `(schema, table) → repo` registry built by discovery; `UndoService` orchestrates undo/history/get with cascade-blocked / already-undone / not-found / recovery-no-path refusals, each carrying a walkable `recovery_action`.
- **Surfaces** — three `system_audit_*` MCP tools + matching CLI commands over the same service; refusals flow through the standard response envelope (parity verified end-to-end).
- **Repo-ification** — new `TransactionNotesRepo` / `TransactionTagsRepo` / `TransactionSplitsRepo`; `transaction_service` composes them. Full-row audit capture; idempotent re-adds emit no noop rows; `clear` emits one `split.remove` per row so each split is individually undoable.
- **Migration + metrics** — `V024` adds `is_undo` / `undoes_operation_id` to `app.audit_log`; `audit_undo_total{outcome}` + reversed-row counter.
- **Specs** — `moneybin-mcp.md`, the capability map, and `CHANGELOG` updated; `data-recovery-contract.md` carries an "as-implemented" note (status stays `in-progress` — PR 4+ remain). The existing `transactions matches undo` migration is intentionally left to PR 5 (it keys on `match_id`, not `operation_id`).

## Test plan

- [ ] `make check` clean; `make test` green (3377 passed, 3 pre-existing skips)
- [ ] Undo round-trip (E2E subprocess): tag a txn → `system audit get` shows it undoable → `system audit undo` reverses it → a second undo is refused with `undo_already_undone`
- [ ] Cascade-block: a later op on the same row blocks `undo(op1)`; undoing the blocker then op1 succeeds
- [ ] Undo-the-undo restores the original rows
- [ ] `match_decisions` (JSON `match_signals`) round-trips through the generic reverser with no override


bsaffel:

PR #222, opened 2026-05-24
Add investments domain specs (M3B) and reconcile references

bsaffel:
## Summary

Initiates the investments domain (milestone **M3B** — the largest competitive moat). Adds the umbrella spec plus its foundation feature spec, and reconciles the rest of the spec tree to point at them. Both new specs are `draft` status — this is design work, not implementation.

The foundation child (Pillars A+B) is scoped so that **realized gain/loss — the 1099-B reconciliation bar — is computed from the transaction ledger alone, needing no price feed.** Price feeds and net-worth integration become clean follow-on children.

## Background

`investment-tracking.md` was the single most-referenced *unwritten* spec in the repo — at least six specs gated on it (Plaid Investments sync, portfolio/holdings reports, investment matching/transfers, investment OFX import, holdings-in-net-worth), yet the file never existed. This PR replaces that placeholder gate with a real umbrella + foundation child, following the Overview-umbrella + feature-children shape every other large domain already uses (Smart Import, Matching, Categorization, Sync, Testing).

Key decisions captured: lots-as-grain with cost-basis method as a stored **election** (FIFO + specific-ID + average-cost as computations over one lot ledger); surrogate `security_id` + resolution chain for identity; append-only price history (not a cache); `currency` column now / FX deferred to M3C; MoneyBin **mirrors** the broker's method rather than enforcing IRS election policy.

## Changes

- **New specs:** `investments-overview.md` (umbrella, four pillars) and `investments-data-model.md` (foundation child — securities dimension, ledger + `type` taxonomy, derived lots/realized-gains/holdings, three-method cost-basis engine, top-level `investments` CLI/MCP surface).
- **Surface placement:** `moneybin-cli.md` promotes `investments` from a placeholder nested under `accounts` to a top-level group (peer of `accounts`/`transactions`/`assets`), migration-table rows updated.
- **Contradiction fix:** `extension-contracts.md` — lot-selection overrides move from the `us_tax` package to core (`app.lot_selections`), since cost basis is core; `us_tax` now *reads* them plus `core.fct_realized_gains` for Schedule D.
- **Reference reconciliation:** 14 `investment-tracking.md` gate references across 12 specs repointed (schema gates → data-model, domain gates → overview); INDEX + roadmap (M3B) entries added.

## Test plan

- [ ] `grep -rn "investment-tracking\.md" docs/ | grep -v worktrees` returns nothing (verified locally).
- [ ] New specs' internal links resolve; the three named planned specs (`investments-price-feeds.md`, `investments-net-worth.md`, `sync-plaid-investments.md`) are intentionally not-yet-created.
- [ ] INDEX.md Investments section and roadmap M3B rows render correctly.
- [ ] Reviewer: sanity-check the core design decisions — method-as-election, security identity (surrogate + resolution chain), top-level `investments` placement, and the `extension-contracts.md` lot-selection move.
- [ ] Pre-commit hooks pass (markdown; no code touched).

### DIFF
--- a/docs/roadmap.md
+++ b/docs/roadmap.md
@@ -163,10 +163,11 @@ The largest competitive moat. M3B does not ship until cost-basis output ties to 
 
 | Area | Status | Notes |
 |---|---|---|
-| Holdings, FIFO lots, realized/unrealized gain/loss | 🗓️ | |
-| Short-term / long-term classification | 🗓️ | |
-| Yahoo + CoinGecko price feeds | 🗓️ | |
-| Plaid Investments product wiring | 🗓️ | Gated on this milestone. |
+| Securities, ledger, lots, realized gain/loss, holdings (cost basis) | 📐 | [`investments-data-model.md`](specs/investments-data-model.md) — foundation child (Pillars A+B) |
+| Short-term / long-term classification | 📐 | Foundation child; oldest-first per-lot under every method |
+| Unrealized gain/loss + Yahoo/CoinGecko price feeds | 🗓️ | Pillar C — `investments-price-feeds.md` (planned) |
+| Holdings in net worth | 🗓️ | Pillar D — `investments-net-worth.md` (planned) |
+| Plaid Investments product wiring | 🗓️ | Gated on the foundation contracts. |
 
 ---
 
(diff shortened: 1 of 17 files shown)


bsaffel:
@claude we good to approve?

PR #223, opened 2026-05-24
Add smart-import confirmation & confidence spec (draft)

bsaffel:
## Summary

New **draft** spec proposing a single "confirm the columns" trust layer shared across all three smart-import channels — tabular, Google Sheets, and PDF. Today each channel runs the same detection engine but handles the result differently (tabular waves medium-confidence through with a log warning, gsheet blocks it, PDF carries a third confidence model inherited from the cut W-2 extractor). This spec unifies them behind **one confidence contract** (normalized `score` + derived `tier`) and **one confirm flow** (`import_files` gate → `import_confirm` commit → `import_preview` inspect), so a user or agent gets the same guarantee regardless of where the file came from: nothing lands unconfirmed on first contact, confirmed layouts reuse silently, and recovery from a wrong guess is one surfaced step away.

Opened **for design review** — see Test plan for the specific decisions I'd like feedback on.

## Background

The umbrella (`smart-import-overview.md`) originally specced an `import_confirm` tool and a propose→review→commit wizard, but the magic-first flow shipped and the confirm half was deferred — the tabular spec's Goal still promises detection "presents it for confirmation," yet only the read-only `import preview` shipped. Meanwhile the PDF spec (#221, lands first) needs `import_confirm` for its bridge rung, where the driving agent hands back a recipe + rows — a genuine data-carrying commit. Rather than let PDF re-derive a PDF-only confirm + confidence model, this spec owns the shared contract and PDF references it.

## Changes

- **`docs/specs/smart-import-confirmation.md`** (new, `draft`) — owns the unified confidence contract (`Confidence` = `score` + `tier` + `flagged` + `missing_required`), the `import_confirm` signature, and the channel-agnostic `resolve_or_confirm` service primitive.
- **Confirm flow:** `import_files` detects a new layout and returns `confirmation_required` (no data loaded); `import_confirm` is the `_commit`-shaped finalize (mapping/accept for tabular/gsheet, recipe+rows for the PDF bridge); `import_preview` stays read-only. gsheet keeps `connect`/`reconnect`, sharing the bands + primitive underneath.
- **Coherence fixes:** `medium` now *gates* on every channel (tabular's current wave-through changes); override becomes *partial-merge* (unifies tabular per-field `--override` with gsheet whole-map `--column-mapping`).
- **Trust model:** first-encounter always confirms; tiered agent autonomy (self-accept `high`) gated behind a calibration bar; revert (`import_revert` / `system_audit_undo`) + re-map surfaced as `actions[]`.
- **`INDEX.md` + `roadmap.md`** — registered (`draft` / 📐 designed).

## Test plan

This is a design spec — "testing" is design review. Reviewer, please weigh in on:

- [ ] **Behavior change:** making tabular `medium`-confidence *gate* (it currently imports with a log warning). Right call for coherence, or too disruptive to the shipped flow?
- [ ] **Confidence unification:** collapsing categorical (tabular/gsheet) + the continuous 0.7-threshold (PDF, ex-W-2) into one `score`+`tier` contract — does the seam hold for both "mapping certainty" and "extraction completeness"?
- [ ] **Mechanism:** `import_files` gate + `import_confirm` commit + `import_preview` inspect — coherent against `surface-design.md` (`_commit` verb, no new shapes)?
- [ ] **Calibration gate:** agent self-accept-on-`high` stays off until ≥0.99 field-exact precision on a fixture corpus. Right bar?
- [ ] Cross-references resolve (note the one intentional forward ref below).

## Related

- Refs #221 — `smart-import-pdf.md`; **this spec's forward reference to it is intentionally dangling until #221 merges.** PDF lands first and will add its reference back to this spec (the agreed cross-spec amendment).
- Refs #219 — `system_audit_undo` is the recovery counterweight this spec leans on.


bsaffel:

PR #231, opened 2026-05-31
Add doctor recipe registry + orphan_app_state audit (M2D PR 4)

bsaffel:
## Summary

- DoctorService failures now ship pre-built, executable `recovery_actions` an agent can dispatch directly — completes the *detect → suggest* half of the M2D data-recovery loop that REC-PR2 plumbed and REC-PR3's `system_audit_undo` consumes.
- Recipe registry (`src/moneybin/audits/recipes/`) maps audit name → `(affected_ids, context) → list[RecoveryAction]`; recipes self-register at package import.
- New `orphan_app_state` audit scans `app.transaction_notes` / `app.transaction_tags` for orphans (transaction_id missing from `core.fct_transactions`) and emits delete/clear actions per orphan, with pending-manual + post-dedup suppression so legitimate user state isn't destroyed.

## What ships

| Audit | Recipe | Confidence |
|---|---|---|
| `orphan_app_state` (new) | per-orphan `transactions_notes_delete` (suggested, non-idempotent across batch) + `transactions_tags_set(tags=[])` (certain, idempotent) | mixed |
| `categorization_coverage` (warn) | `transactions_categorize_run(methods=["rules","merchants"])` (cascade can apply 0 rows when no rules/merchants match — verify after) | suggested |
| `dedup_reconciliation` (fail) | `refresh_run()` full cascade + `system_doctor(full=True)` (match alone updates app.match_decisions but does not rebuild core) | suggested |

Audits without a registered recipe leave `recovery_actions=None` rather than emit fabricated "investigate manually" hints — per-domain retrofit is PR 9a–f.

## Wiring

- **Single seam:** `DoctorService.run_all` walks each `InvariantResult` through `_apply_recipe`. Per-`_run_*` methods stay unaware of the registry. Recipe exceptions are caught per-invariant so one bad recipe can't take down the report.
- **`InvariantResultPayload`** gains `recovery_actions: list[RecoveryActionPayload]` (mirrors `SystemAuditHistoryEntryPayload` precedent — required list, possibly empty).
- **MCP `system_doctor` adapter** + **CLI `system doctor`** text and JSON renderers both carry the new field. CLI text format: `💡 [confidence] tool(arguments) — rationale` indented under each failing/warning invariant; capped at 5 actions per invariant with a "… and N more" trailer.
- **Migration V026** adds `transaction_id` to `raw.manual_transactions` so the orphan audit can suppress false-positives on pending-but-not-yet-materialized rows. Nested `NOT EXISTS(prep.int_transactions__matched)` discriminator lets the post-matcher deduped case still surface real orphans.

## Notable design choices recorded in the spec

- `affected_ids` for `orphan_app_state` use prefixes (`note:<id>` / `tag:<txn_id>`) so the recipe dispatches by entity type without re-querying the DB.
- `transactions_notes_delete` is cited in single-id form for now; PR 8 will add a list form and the recipe will upgrade to `confidence="certain", idempotent=True`.
- Spec is the canonical record — see the new "PR 4 — as implemented" subsection in `docs/specs/data-recovery-contract.md`. Spec status stays `in-progress` (PR 5+ remain).

## Test plan

- [x] `make check test` — full unit suite passes
- [x] `make test-scenarios` — 45 scenario tests pass (data-shape change required this)
- [x] `uv run pyright` on all changed files — 0 errors
- [x] Round-trip-executable test — every recipe's `tool` resolves to a registered MCP tool, `arguments` bind to its signature, and Literal-typed values are members of their literal (catches `methods=["rule"]`-style typos)
- [x] Integration tests against real encrypted DuckDB: orphan note/tag flagging, pending-manual suppression, deduped-after-matching reporting, lockstep hash between migration and service
- [x] MCP envelope shape test: `system_doctor` response carries `recovery_actions` per invariant with the right confidence values
- [x] CLI render tests: text + JSON both surface the actions; text caps at 5 with trailer


bsaffel:

PR #240, opened 2026-06-08
Add writer coordination for concurrent database access

bsaffel:
## Summary

MoneyBin opens a short-lived, purpose-declared DuckDB connection per operation (ADR-010). DuckDB permits only one writer to a file cross-process, so when several MoneyBin processes write concurrently — the MCP server, a Plaid sync, the transform pipeline — a second writer could race at DuckDB's ATTACH layer and surface a raw `IOException` instead of a clean, retryable error.

This PR adds a per-profile **write critical-section lock** in front of every write-mode open, retried under a single shared deadline; a **checkpoint at durable boundaries** so a crash after a migration or transform can't lose committed state; **multi-process test coverage**; and a **`system_status` view of who currently holds the database**. Writer contention now always produces a classified, retryable `DatabaseLockError` envelope — never a raw DuckDB exception.

## Background

Two near-term capabilities break the old single-writer assumption: a web UI that accepts categorization edits while the MCP server runs, and a Plaid sync that writes transaction batches out-of-process. The cross-process lock matrix was re-confirmed empirically on DuckDB 1.5.3 (read+read coexist; every combination involving a writer conflicts).

The branch went through a thorough review pass before this PR. That surfaced correctness issues across three areas — the lock primitive (the holder-metadata write was swapping the lock file's inode and defeating the lock; reentrancy was keyed too coarsely), the `Database` lifetime (lock/connection teardown ordering and a construction path that could leak a connection), and the recovery-action envelope — all fixed here with regression tests. Manual two-process verification then caught a further regression where the lock ran before the profile directory existed; also fixed.

## Changes

Reviewer orientation — where to focus:

- **Lock primitive (`src/moneybin/db_lock/`)** — `write_lock` uses `fcntl` with exponential-backoff retry against a deadline shared with the ATTACH loop, so the end-to-end writer wait stays under the 10-second policy ceiling. Holder metadata is written in place through the held descriptor (never a temp-file rename); reentrancy is keyed on `(pid, thread_id)`.
- **Connection lifetime (`database.py`)** — the lock is held for the full lifetime of the returned `Database`, not just during ATTACH. `close()` tears down the DuckDB connection before releasing the lock; `__init__` uses a construct-or-rollback guard so a failed open never leaks a connection.
- **Durability (`Database.checkpoint`)** — emits `CHECKPOINT` + observability at named boundaries (post-migration, post-transform). A checkpoint failure is a durability hint, not a correctness signal — it never flips an applied transform to "failed".
- **Diagnostics (`system_status`)** — a `database_connections` block reports the live writer and reading processes. A writer is reported only when a process actually holds the lock (a `LOCK_SH` probe), so a persisted-but-released lock file never shows a phantom holder. `DatabaseLockError` carries a structured recovery action pointing the agent at this view.
- **Coverage** — six multi-process scenarios spawn real subprocesses against an encrypted DuckDB file and assert that contention always yields a classified envelope; unit + integration regression tests pin each fixed bug.

## Test plan

- [ ] `make check` — 0 errors (3 pre-existing reportlab import warnings only)
- [ ] `make test` — 3832 passed, 4 expected skips
- [ ] `make test-scenarios` — 51 passed, including all 6 writer-coordination scenarios
- [ ] Manual: race two concurrent `moneybin transform apply` against one profile — both succeed, the lock metadata shows `operation_type=transform_apply` on each holder while held, both emit a `post_transform` checkpoint, and no raw `IOException` / "Could not set lock" appears in either process
- [ ] Reviewer to verify: the `LOCK_SH` probe distinguishes a live writer from a stale lock file; `close()` releases the connection before the lock; the shared deadline keeps total writer wait under the 10 s ceiling


bsaffel:
Re: the round-6 "scenario tests bypass `write_lock`" finding — the `write_lock` cross-process invariant **is** tested end-to-end. `tests/scenarios/test_writer_coordination.py` (6 scenarios) drives real subprocesses through `get_database(read_only=False)` → `write_lock` against a real encrypted DuckDB (see `tests/scenarios/_lock_workers/holder.py` + `contender.py`: two-writer serialization, writer-timeout envelope, post-checkpoint reopen, etc.).

`tests/e2e/test_concurrent_access.py` is **pre-existing** (on `main`, unchanged by this PR) and *deliberately* opens `Database(...)` directly to probe DuckDB's native ATTACH-layer locking — the layer `write_lock` sits on top of. The two files cover the two layers. I've added a docstring cross-reference between them (17cb4483) to make that explicit.

The `ps`-subprocess-under-`LOCK_EX` finding from the same review is fixed in 17cb4483 — `_process_command` now runs before the `flock` acquire, so only the fast in-place metadata write happens under the lock.

PR #258, opened 2026-06-18
Account identity: bind-first + exporter/institution split, with de-shortcut scenario (M1S.8/.9)

bsaffel:
Builds on #257 (M1S.7 capture layer, merged). With last4 captured, the cross-source matcher *fires*; this increment makes the surrounding account model **bind-first** and splits the **exporter axis from per-account institution**, and closes a long-standing scenario shortcut that hid the original bug.

## M1S.8 — bind-first / format-account decoupling

- **A saved tabular format carries columns only, never an account binding.** The auto-saved `institution_name` is now the resolved (filename/format) institution or `"unknown"` — never the per-account `--account-name` label (bug #5). A second structurally-identical CSV imported with a *different* explicit account name lands on a distinct account (regression-pinned).
- **A renamed aggregator account re-associates instead of duplicating.** An aggregator account renamed `Everyday Spending (...0001)` → `Fun Money (...0001)` (same last4, new slug) misses the exact-slug strong-ref and re-associates onto the original via a *pending* `institution_last4` proposal — never a duplicate mint or silent merge (pinned; resolver doc'd).

## M1S.9 — exporter ≠ institution

- **Per-account institution from row data for multi-account formats.** A Tiller-style export with a per-row Institution column now records *each account's own* institution (both the resolver input and `raw → dim`), never the single shared exporter/tool name. Single-account behavior is unchanged. (The Monarch case — institution embedded in the account *label* — needs a label→institution parser and is a later increment.)
- Names the exporter/format axis (`source_origin`) distinct from per-account institution in code + the `app.tabular_formats` schema comment.

## Scenario de-shortcut (the root-cause fix)

The original cross-source bug stayed green for ~6 days because every cross-source scenario **forced** the account link (`account_bindings` / a shared `account:` fixture label) instead of exercising the matcher. This adds an **end-to-end scenario** that imports the `.qfx` + `.csv` twin **without** any binding, runs matching, **accepts the proposal through the real review-queue accept**, and asserts the twins collapse to one gold set — proving the full `match → accept → collapse` chain a user actually relies on. A new testing rule, **"No Shortcuts: Exercise the Real Mechanism,"** codifies this for future scenarios.

## Test plan

- `make check test` → 4356 passed
- account-identity + capture-contract scenarios → 10 passed, incl. the new end-to-end `match → accept → collapse`
- Final whole-branch review (clean; the two findings it raised are fixed in the last commit)

### DIFF
--- a/.claude/rules/testing.md
+++ b/.claude/rules/testing.md
@@ -194,3 +194,49 @@ When authoring or modifying a scenario:
 5. **Negative expectations are required where applicable.** If a scenario asserts "these N records should match," it must also include cases that should *not* match. Otherwise the test only catches under-matching, not over-matching.
 
 This rule applies to YAML scenario expectations, pytest assertions in `tests/scenarios/`, and any future bug-report-driven scenario. See [`docs/specs/testing-scenario-comprehensive.md`](../../docs/specs/testing-scenario-comprehensive.md) for the full taxonomy and contributor recipe.
+
+## No Shortcuts: Exercise the Real Mechanism
+
+A test or scenario must reach its asserted state through the **same mechanism a
+real user or agent would use**. Never pre-wire the end state to make an
+assertion pass when the mechanism that produces that state is the thing under
+test. A green test that bypasses the mechanism proves nothing about the
+mechanism — it hides bugs.
+
+**The canonical failure this rule exists to prevent:** cross-source account
+linking was silently broken in the wild through ~6 days of development because
+the only cross-source scenario *forced* the CSV→OFX account link — via
+`account_bindings` (the import path) and a shared `account:` label on the
+fixture pair (the `load_fixtures` path) — instead of letting the account matcher
+fire. The scenario stayed green the entire time while automatic matching never
+once worked. See `account-identity-resolution.md` (Decision 8).
+
+**When the mechanism IS what the scenario validates, never substitute for it:**
+
+- **Account identity / matching:** don't force the link with `account_bindings`,
+  a shared `account:` fixture label, a pinned `account_id`/`explicit_account_id`,
+  `force_standalone`, or a direct `INSERT INTO app.account_links`. Import the raw
+  twins and let `AccountResolver` produce the proposal.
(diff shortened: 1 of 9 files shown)


bsaffel:
Thanks for the review. The 🔵 nit is fixed in 1bc3c677. After verifying the other two against the code and git history, I believe both are false positives:

**🔴 `import_service.py:1426`** — This line is the body of the `elif (resolved.is_multi_account and acct_name_col and acct_name_col in df.columns):` branch (line 1389), so `resolved.is_multi_account` is **always `True`** here. A single-account file without `--account-name`/`--account-id` raises at the `else` on line 1434 and never reaches 1426 — so a single-account bank CSV like `wells_fargo_export.csv` cannot hit this path. The suggested guard `... if resolved.is_multi_account else institution` is therefore a no-op (the condition is always true in this branch). And making it meaningful — falling back to `institution` — would re-stamp the *exporter* identity (Monarch/Tiller) onto every account, which is exactly the Decision-8 conflation that `2d0aec0a` intentionally removed (`-institution=institution` → `+institution=multi_acct_inst.get(native_key)`). For a multi-account aggregator export with no Institution column, `None` is the correct per-account value.

**🟡 institution-resolved matching coverage** — The premise that explicit `account_name` bypasses institution-based matching isn't accurate. `test_csv_twin_matches_accepts_and_collapses_end_to_end` imports the CSV twin *with* `account_name="WF Checking (...1111)"` and asserts (line 277) that the matcher fires an `institution_last4` proposal, with the institution resolved from the filename. The explicit account name only sets the source-native key; the institution still flows into `AccountResolver` and drives the weak-candidate pass. The specific test suggested (single-account CSV *without* `account_name`) isn't constructible — that path raises by design.

One real, separate nuance the finding brushed near: a *single-institution* multi-account export with no Institution column would lose the filename institution under `None`. That's a minor future refinement — it can't be addressed by a blanket fallback without reintroducing the aggregator bug — and is out of scope for this PR.

PR #279, opened 2026-06-28
Scope sync identity and tokens to the active profile

bsaffel:
## What

The sync client now authenticates as a distinct broker user per profile, instead of all profiles sharing one token slot and one server identity.

- **Stable opaque profile id** — `get_or_create_profile_id()` lazily generates a truncated-UUID4 id under `<profile_dir>/profile_id` (profiles created before this get one on first sync). Opaque on purpose: no profile name reaches the broker subject.
- **Per-profile token storage** — `SyncClient` namespaces its keychain entries (and fallback token file) by `profile_id`, so switching profiles uses the right token and one profile can't reuse another's identity.
- **Login** — `SyncClient` forwards `profile_id` at `/auth/device/token`; `_build_sync_client` derives the active profile from the resolved DB path and passes its id.

## Why

A "profile" was purely local (its own DB / config / keychain) but invisible to the broker — every profile minted the same `self-hosted-user` subject, so sandbox items linked under one profile followed you into another. This makes each profile a distinct broker identity.

## Coordination

Pairs with the broker change (moneybin-sync `feat/per-profile-identity-and-env-guard`), which must merge/deploy first — the client sends `profile_id` only when set (always, in practice), and a broker without the matching change would reject the extra field.

## Tests

New tests for the profile-id helper (generate + stability) and `SyncClient` (per-profile keychain isolation + login forwarding the id). Full moneybin unit suite green (4413); ruff + pyright clean.

### DIFF
--- a/docs/reference/server-api-contract.md
+++ b/docs/reference/server-api-contract.md
@@ -89,8 +89,13 @@ The client reads `user_code` (display), `verification_uri_complete` (open in bro
 **Request:**
 
 ```json
-{ "device_code": "Ag_EE..." }
+{ "device_code": "Ag_EE...", "profile_id": "ab12cd34ef56" }
 ```
+
+`profile_id` is optional — an opaque, stable per-profile identifier the client
+sends so each local profile maps to a distinct broker identity. Omit it for the
+legacy single-identity behavior. When present it must match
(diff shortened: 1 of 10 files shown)


bsaffel:

bsaffel:
@claude you aren't leaving any comments, please approve the PR unless you have a reason not to

PR #284, opened 2026-06-29
Resolve transaction merchants by Plaid merchant_entity_id (M1T)

bsaffel:
## Summary

Resolves a transaction's merchant by Plaid's stable `merchant_entity_id` **before** name matching — the merchant twin of M1S account-identity resolution. One real merchant collapses to one canonical `merchant_id` even when its description varies across transactions.

## What's added

- **Schema** — two `app.*` tables: `merchant_links` (provider-id → canonical `merchant_id` binding; provider-neutral `(source_type, ref_kind, ref_value)`, N:1) and `merchant_link_decisions` (fuzzy-match review queue), written through audit-paired repos (Invariant 10). V031 catch-up migration; privacy classification (opaque ids `RECORD_ID`, merchant names `MERCHANT_NAME`).
- **Resolver** — `MerchantResolver` adopt-or-mint ladder at categorization time: adopt a bound id → auto-bind an exact name match → propose a fuzzy match to review → mint a `created_by='plaid'` merchant. Wired into both categorization write paths.
- **Pipeline** — `merchant_entity_id` (and an entity-paired `merchant_entity_source_type`) carried through `prep.int_transactions__merged`, deliberately **not** into `core.fct_transactions`: consumers read canonical `merchant_id`/`merchant_name` from core and call the link registry for the provider id.
- **Backfill** — `harvest()` records existing categorizations as bindings with zero review; multi-merchant conflicts route to the review queue.
- **Surfaces** — `merchants links {pending,set,history,run}` CLI + `merchants_links_{pending,set,history,run}` MCP tools; the top-level `review` orientation tool gains a fourth (merchant-links) queue. Plus metrics.

## Notable decisions

- Bindings key on the **entity-issuing provider**, not the transaction's merge-winning source — so a Plaid entity id riding an OFX+Plaid-deduped transaction still binds under Plaid, preventing a cross-source merchant split regardless of `source_priority` config.
- `undo` for link decisions is deferred to M1L's audit-undo consumer, matching `accounts links`.

## Testing

Full gate green: unit (4501) + scenarios (75) + e2e (313), including an OFX+Plaid same-entity no-split scenario that drives the real pipeline end to end.

Spec: `docs/specs/merchant-entity-resolution.md`.


bsaffel:
@claude where's your review? do you approve?

bsaffel:
## Review round close-out — addressing the remaining items

Thanks for the thorough passes. The one actionable item from the latest review is now fixed, and here's the disposition of the remaining open threads so we can converge.

### Fixed this push

- **`review --type merchant-links`** — the `--type` filter now accepts `merchant-links` (added to `_VALID_TYPES`, both the JSON and text per-queue branches, the two `--help` strings, and `moneybin-cli.md`), so `moneybin review --type merchant-links --status` shows just that queue instead of erroring. Tests added.

### Open 🟡 threads — why we believe they're resolved

- **`merchant_entity_id` ARG_MIN sentinel (`int_transactions__merged.sql:51`).** This now uses `COALESCE(sp.priority, 2147483646)`, matching `merchant_entity_source_type`. To be explicit about the underlying concern: this is a **no-op** for `merchant_entity_id`, because DuckDB's `arg_min(arg, val)` **ignores rows where `arg` is NULL**. We verified empirically — with two tied rows `('PLAID_E', 2147483647)` and `(NULL, 2147483647)`, `arg_min(merchant_entity_id, k)` returns `'PLAID_E'` every time, never NULL. So an entity-bearing member always wins the tie and `merchant_entity_id` cannot resolve to NULL when one exists — the column can't drop the id. (That asymmetry is also *why* the original change only added the sentinel to `merchant_entity_source_type`, whose `arg` is non-NULL for every member.) We've made the two ARG_MINs use the same sentinel anyway for readability. The trailing comment documents this.

- **Rung-3 proposes the single best candidate (`merchant_resolver.py`).** This is a deliberate scope decision, now reflected in the spec (Decision 3, rung 3 reads "the single best name-match candidate"). `match_merchants()` is single-best by construction; binding an entity id to a *different* existing merchant than the proposed candidate is the **merchant-merge** surface, which this spec explicitly puts out of scope. The N-candidate review-queue data model (`PendingMerchantLinkGroup.candidates`, `set()` sibling auto-reject) already exists as forward-compatible substrate for when that surface lands.

### 🔵 Nits — tracked as follow-ups (non-blocking)

- **Resolver rung-4 mints not counted in `merchants_created`.** Categorization is correct; only the count is understated (resolver mints are `created_by='plaid'`, distinct from the exemplar `'ai'` path). Follow-up.
- **Tool registration order (`merchants_links_run` before `_set`).** Cosmetic — the full tool surface is visible at connect regardless of registration order. Follow-up.
- **`V031` `decided_by` CHECK asymmetry comment.** The asymmetry (`decisions`: `auto`/`user`; `links`: `auto`/`user`/`system`) is intentional and documented in the spec. We're deliberately *not* editing `V031` because amending a shipped migration changes its checksum (flagged by `check_drift`). Noting the rationale here instead.

The spec is now marked `implemented`. **With the above, we believe this is ready — could you re-review and approve if you agree?** The three nits are tracked as follow-ups.

bsaffel:

PR #300, opened 2026-07-06
Add investments data model & cost-basis engine (M1J.1)

bsaffel:
## Summary

Ships the investment data model & cost-basis engine (M1J.1) — the foundation
child of the M1J investments keystone. A manually-maintained securities
catalog, an investment-transaction ledger (14-type closed taxonomy), and a
pure-Python cost-basis engine (FIFO, HIFO, specific-ID, average-cost) derive
tax lots, realized gain/loss (1099-B reconciliation grain), and current
holdings — surfaced via a top-level `investments` CLI group and matching MCP
tools. This closes the contracts ~6 gated specs (Plaid Investments sync,
portfolio/holdings reports, investment matching/transfers, investment OFX,
holdings-in-net-worth) have been waiting on.

The final push includes a high-effort pre-push review pass that surfaced and
fixed 10 real correctness bugs (see Changes), closed a zero-E2E-coverage gap
for the whole `investments` CLI surface, added two regression tests pinning
accepted-but-previously-untested v1 behavior, and confirmed 2 other findings
were already-documented, deliberate v1 scope decisions rather than bugs. One
further finding (currency mixing in cost-basis grouping) is accepted as v1
scope and tracked as follow-up work — see Background.

## Background

Cost basis is computed once, per `(account, security)` group, from the full
ledger — never rewritten historically. Corporate actions (`split`,
`return_of_capital`) mutate open lots in place at their trade-date position;
acquisitions (`buy`, `transfer_in`, `reinvest`) open lots; disposals (`sell`,
`transfer_out`) consume them per the elected method. The engine never raises
on incomplete data — an oversold disposal or a `transfer_in` with unknown
basis opens/realizes at zero basis and sets `basis_incomplete` rather than
inventing a number or blocking the rebuild.

Two review findings turned out to be pre-existing, intentional v1
limitations rather than new bugs, confirmed by reading their pinned tests
and docstrings before touching anything:
- The return-of-capital clamp can drop part of a reduction across
  uneven-basis lots (`cost_basis.py`'s own docstring calls this out as "a
  known simplification... revisit in a follow-up," pinned by
  `test_return_of_capital_clamp_overflow_is_dropped_v1`).
- `core.fct_investment_lots` and `core.fct_realized_gains` each independently
  re-run the engine per SQLMesh refresh — an acknowledged "cheap at
  personal-finance scale" tradeoff, not an oversight.

Both are left as-is; there's one remaining opportunistic cleanup tracked as
follow-up work (a duplicated write-path pair in `InvestmentService`).

A third finding, raised post-approval, is accepted as v1 scope rather than
fixed here: `compute_lots_and_gains` groups lots/pools by `(account_id,
security_id)` only, not `currency_code`. `record_event`'s per-transaction
currency is never cross-checked against the security's registered currency,
so a transaction recorded with a mismatched currency would silently mix
amounts across currencies in the same lot/pool math. This is consistent with
the documented v1 scope (currency is tracked now; FX conversion is deferred
to a later multi-currency milestone per `investments-data-model.md`) and
low-risk today — a security's currency is consistent per source in practice.
Tracked as follow-up work to close alongside that milestone.

## Changes

- **Data model.** `core.dim_securities` (surrogate id + CUSIP/ISIN/ticker/name
  resolution chain), `core.fct_investment_transactions` ledger, migration
  V034. `app.lot_selections` for specific-ID overrides.
- **Cost-basis engine** (`investments/cost_basis.py`). FIFO, HIFO,
  specific-ID, and average-cost as elections over one lot grain; splits and
  return-of-capital as in-place corporate actions; reinvest pairing;
  oversold/incomplete-basis handling with penny-conserved rounding.
- **Service + repos.** `InvestmentService` read/write paths, `SecuritiesRepo`,
  `LotSelectionsRepo` (Invariant 10 — all `app.*` mutation through a repo).
- **Surfaces.** Top-level `investments` CLI group (replacing the earlier
  `accounts investments` placeholder) and matching
  `investments`/`investments_holdings`/`investments_lots`/`investments_gains`/
  `investments_securities` read + `investments_record`/
  `investments_securities_set`/`investments_lots_select` write MCP tools.
- **Pre-push review fixes:**
  - `upsert_security`/`set_security` now hard-validate `cost_basis_method`/
    `security_type` against their closed vocabularies before the write,
    matching the `AccountService.settings_update` pattern from this same PR
    — previously an invalid value reached the DB and raised a raw
    `duckdb.ConstraintException` instead of a clean `UserError`.
  - `ImportService.allocate_import_log`'s synthetic `source_file` key is now
    namespaced by `format_name`, so `revert()`'s superseded-batch lookup
    can't cross-match a manual cash-transaction batch with a manual
    investment batch (both share `source_type="manual"`).
  - `core.fct_investment_lots` gains a `basis_incomplete` column: a
    `transfer_in` with no supplied basis (real-world ACATS transfers of
    uncovered securities often arrive this way) now opens a flagged
    zero-basis lot instead of one indistinguishable from a legitimately
    zero-cost lot, and the flag carries through to any realized gain a later
    disposal draws from it.
  - `select_lots` now rejects a specific-ID selection naming a lot acquired
    *after* the disposal's trade date, not just one from the wrong
    `(account_id, security_id)`. The engine's chronological replay only adds
    a lot to a group's open-lot list once its acquisition event has been
    processed, so a lot acquired later wasn't in `_consumption_plan`'s
    `by_lot_id` at replay time — the selection silently fell back to FIFO
    while the CLI/MCP still reported the override as applied.
  - A `reinvest`'s income leg no longer inflates taxable dividend/interest/
    capital-gain income by the transaction fee. `amount` is fee-inclusive by
    convention (the full cash redeployed), so negating it directly for the
    income leg baked the fee into reported income (a $150.00 dividend
    reinvested with a $4.95 fee recorded $154.95 of income); the fee is now
    added back before negating.
  - `investments_securities_set`'s update path now rejects an attempt to
    change `security_type` instead of silently dropping it while returning
    an `ok` envelope — `security_type` is immutable post-creation (as its
    own docstring already stated), matching the CLI `set` command, which
    doesn't expose the flag on update at all.
  - `select_lots` now also rejects a selection already exhausted (fully or
    partially) by an *earlier* disposal, not just one acquired too late.
    `core.fct_investment_lots` retains closed lots, and the prior check only
    validated position/date — it never checked consumption history, so a
    lot already drawn down by an earlier-dated disposal passed validation
    and silently fell back to FIFO for the unavailable quantity at replay
    time. The check now sums each lot's consumption from
    `core.fct_realized_gains` by strictly-earlier disposals and rejects a
    selection that exceeds what's actually left. `investments_lots_select`
    also gains a `refresh_run`/`investments_gains` actions hint, matching
    `investments_record`'s pattern.
  - That same consumption check is now blind spot-free for `transfer_out`:
    `transfer_out` consumes lots but realizes no gain, so it never writes a
    `core.fct_realized_gains` row, and the prior version above only summed
    that table — an earlier `transfer_out`'s draw-down was invisible to it.
    The check now derives availability from each lot's current
    `remaining_quantity` (which nets out every disposal type) and adds back
    only what a non-earlier sell drew via `fct_realized_gains`, closing the
    gap for both disposal types.
  - `investments add`, `investments securities add`, and `securities set`
    built their JSON success envelope from a bare dict with a hardcoded
    `sensitivity="low"` instead of the already-imported typed payload
    (`InvestmentRecordPayload`/`InvestmentSecuritySetPayload`) — the privacy
    audit trail logged `classes_returned=[]` instead of the real data class,
    and any future redaction transform would silently be skipped on the CLI
    while still firing on the equivalent MCP tool. Same anti-pattern already
    fixed once in this PR for `investments_lots_select`; now fixed at all
    three remaining call sites.
  - `moneybin doctor`'s Invariant-10 audit-coverage sweep didn't cover
    `app.securities`/`app.lot_selections` — every other repo-wrapped table
    has a registered coverage check, so a raw bypass write to either new
    table would go undetected. `lot_selections` keys the check on
    `investment_transaction_id` alone (not the composite PK), matching
    `LotSelectionsRepo`'s collection-shaped audit granularity (one
    `lot_selections.set` row per disposal, not per `(txn_id, lot_id)` pair).
- **Zero E2E coverage for the investments CLI surface.** The 9-subcommand
  `investments` group (`add`, `list`, `holdings`, `gains`, `lots list`,
  `lots select`, `securities list`, `securities add`, `securities set`) had
  only in-process `CliRunner` unit tests — a direct violation of this
  project's "every CLI command must have an E2E subprocess test" rule. Added
  a full lifecycle test in `test_e2e_workflows.py`: mint an account via the
  existing OFX fixture (investments are import-first for the account,
  ledger-first for transactions), add a security, record a buy and a sell,
  apply transforms, exercise the specific-ID lot-selection override, and
  drive every read command.
- **Two accepted-behavior regression tests, no code change.** (1) The
  same-day, same-type-bucket disposal tie-break (`_SAME_DAY_TYPE_ORDER`'s
  `investment_transaction_id` fallback) had no test proving it's
  deterministic and stable across input order. (2) Changing a security's
  `cost_basis_method` after a disposal is realized silently recomputes that
  disposal's cost basis on the next refresh (the `FULL`-kind models re-derive
  the entire history from the *current* method) — accepted v1 scope (no IRS
  election lock-in enforcement), now pinned by a test and called out in the
  CLI `--method` help text and the `investments_securities_set` MCP
  docstring so it isn't an undocumented surprise.

## Test plan

- [x] `make check test` — 4950 unit tests pass
- [x] `make test-scenarios` — 77 scenarios pass, including
      `test_investments_1099b_reconciliation` (hand-labeled full-tax-year
      fixture) and `test_investments_persona_correctness`
- [x] `make test-e2e` — 327 E2E subprocess tests pass, including the new
      full-lifecycle `investments` workflow test
- [x] New tests added for each pre-push fix: engine-level (`basis_incomplete`
      propagation from acquisition to disposal, same-day tie-break), service-
      level (validation, revert-collision), CLI/MCP-level (JSON + text
      output, audit-trail `classes_returned`), doctor-level (audit-coverage
      pass/bypass for both new tables), and real SQLMesh-pipeline integration
      tests (`test_investment_models_transform.py`)
- [x] `uv run pyright` clean on all touched files


bsaffel:
@claude The PR description now explicitly accepts the currency-grouping deferral (Codex's `cost_basis.py:731` finding) as v1 scope, matching the existing pattern for the other two accepted v1-limitation callouts in Background — no code change accompanies this, just the deferral writeup. Please re-review.

bsaffel:
@claude The last automated review run on this commit (df6f14a1) completed without posting any review, comment, or verdict. Please re-review the current HEAD — commit df6f14a1 fixes the reinvest fee-inflated-income bug and the silent security_type no-op on update that the prior review (at 77c3288f) flagged.

PR #315, opened 2026-07-11
Fix install and onboarding paths that misdirect or dead-end users

bsaffel:
Two first-contact surfaces were shipping wrong output. Both are on the path a stranger hits before MoneyBin has done anything useful for them.

## `profile create` could not repair a half-made profile

A profile directory with no `config.yaml` — left by a bare `moneybin db init`, a hand `mkdir`, or an interrupted delete — was a dead end. `create()` raised `ProfileExistsError` off `mkdir(exist_ok=False)`, i.e. off the **directory** rather than off a registered profile, so nothing could ever finish it: `profile create` refused, `profile list` hid it (it filters on `config.yaml`), and it never got an import inbox.

`ProfileExistsError` now means *"a registered profile exists"*. `create()` adopts an unregistered directory and completes it in place:

- an existing database is never touched, and an adopted directory is never rolled back (`rmtree` there would destroy the data the user is trying to recover);
- the directory is chmod'd to `0o700` — whatever created it did so at the ambient umask (typically `0o755`), and the encrypted database and privacy log are about to live behind it;
- `config.yaml` is written **last**, as the commit marker. A `create` that dies on a locked keychain now leaves the profile unregistered and therefore retryable, instead of visible-but-broken and refused by its own retry hint.

`DemoService` no longer infers *"this profile already existed"* from that exception — it asks the filesystem. **That inference was a live data-loss path**: with the contract changed and demo left alone, `moneybin demo` walks straight past its data-safety guard and rebuilds a `db init`'d profile named `demo` holding real data. The regression test fails with `DID NOT RAISE` if it ever returns.

First-run guidance follows: an unregistered profile is pointed at `profile create` (which now repairs it) rather than `db init`. A registered profile missing only its database still gets `db init`.

## `mcp install --client chatgpt-desktop` never actually installed anything

It printed a config snippet and told the user to "choose the local/stdio option" in ChatGPT's Connectors UI, calling that "the supported, authenticated path" — while writing **no config at all**. Following the instructions got you nowhere.

The fix, after getting it wrong once (see below): the ChatGPT desktop app **hosts Codex and shares its MCP configuration**. Per OpenAI's docs, *"The ChatGPT desktop app, Codex CLI, and IDE extension support MCP servers and share MCP configuration for the same Codex host"*, and the app exposes **Settings → MCP servers → Add server → STDIO**. That config is `~/.codex/config.toml` — the same file `--client codex` already writes.

So `chatgpt-desktop` now performs a real install into the shared Codex config (one install serves the ChatGPT desktop app, the Codex CLI, and the IDE extension), names the restart step, and warns that ChatGPT on the **web** genuinely cannot see a local server — that surface is remote-only and needs M3D.

> **Correction, recorded deliberately.** An earlier commit on this branch made `chatgpt-desktop` *hard-refuse*, on the claim that no ChatGPT surface can reach a local server. That was false. I had checked OpenAI's **API** MCP docs — which are remote-only because they describe the Responses tool, not the desktop client — and treated that as settling the client surface. The `chatgpt-codex-connector` bot flagged this and I incorrectly argued back. The refusal was a *fresh misdirection of exactly the kind this PR exists to remove*, and strictly worse than the bug it replaced, because it blocked an install that works. Fixed in 925c923b.

Snippet hardening (unaffected by the above):

- **absolute `uv` path** — GUI-launched macOS clients don't inherit the shell `PATH`, so a bare `uv` resolved to nothing and the server died at launch;
- **Codex `startup_timeout_sec = 30`** — a cold `uv run` takes 3–15s against a 10s default, making the *first* connection the one most likely to time out;
- **advisory output moved to stderr** — the guide promises `--print` "emits the exact bytes the command would write", but notes and hints were landing on stdout. `mcp install --print | jq` now works; a test parses stdout as JSON to keep it that way.

Gemini CLI's `trust: true` bypasses **all** tool-call confirmations. MoneyBin doesn't set it — our surface includes write tools — and now says so, rather than leaving the omission to read as an oversight.

## MoneyBin exceeds Windsurf's tool cap (found while verifying)

Verifying the connect-time tool count turned up a shipped defect rather than a clean bill. MoneyBin exposes **102 tools with none hidden** (progressive disclosure was retired), and Cascade holds a hard maximum of **100 tools at any one time, across all your MCP servers**. Windsurf gives no signal when it drops the overflow — it simply behaves as though MoneyBin can't do things it can.

`mcp install --client windsurf` now warns about this, the guide documents it, and a test pins the visible count against the cap so growth is a conscious act. The counts live in `moneybin/mcp/surface.py` (plain constants, no FastMCP import, so the CLI cold-start path stays clean) and the test asserts them against the live registry, so the number the CLI prints can't drift from what we ship.

**Getting under 100 is a surface decision, not this PR's** — it means choosing which tools to hide, which deserves its own discussion.

The Claude Desktop guide also now leads with `.mcpb` extensions (config JSON is legacy-but-supported; MoneyBin's own bundle is still M3B) and documents two failures that read as bugs: Cowork's *remote* sessions can never see a local MCP server, and managed-org flags can disable local MCP outright.

## Note for the AI-client-compatibility spec

`docs/specs/ai-client-compatibility.md` §"Blessed-path corrections" item 2 rests on the same false premise I did — that all ChatGPT surfaces are remote-only. **That should be corrected before the spec merges**: the desktop app is not remote-only; ChatGPT web/mobile are.

## Verification

`make check test` green (5125 passed), e2e and integration suites green. Every path driven by hand through the real CLI — which is how the "database left untouched" lie, the half-registered failure state, and the `--print` stdout pollution were each caught. None of them showed up in tests until the command was actually run.


bsaffel:
Adjudicating the open `chatgpt-codex-connector` thread (that ChatGPT Desktop shares Codex's local MCP path), since the review flagged it as unverified:

**The bot is wrong — it conflates two different OpenAI products.**

- **Codex** (CLI / Desktop / IDE) speaks local stdio MCP via `~/.codex/config.toml`. MoneyBin already supports it — that is the existing `--client codex`, untouched by this PR.
- **ChatGPT** (web / desktop / mobile) reaches MCP only through remote connectors. Per OpenAI's own MCP documentation, connectors are configured with a `server_url` HTTPS endpoint ("Remote MCP servers can be used to connect models over the Internet…"); there is no stdio/local-command transport documented for ChatGPT's developer mode or custom connectors.

So the two clients are genuinely different surfaces, and both are now handled correctly: `codex` installs, `chatgpt-desktop` refuses and explains. Nothing to change here.

Separately, the 🟡 on the missing Windsurf install-time warning is fixed in 3cc86556 — `mcp install --client windsurf` now warns on stderr that MoneyBin's 102 tools exceed Cascade's 100-tool ceiling and points at how to get under it. The counts moved to `moneybin/mcp/surface.py` (plain constants, no FastMCP import, so the CLI cold-start path stays clean) and the budget test asserts them against the live registry, so the number the CLI prints can't drift from what we ship.

bsaffel:
**Retracting my previous comment — I was wrong and the `chatgpt-codex-connector` bot was right.**

I adjudicated that thread against the bot without checking the page it was actually pointing at. I checked `developers.openai.com/api/docs/mcp` (the *API* Responses-tool docs, which are remote-only) and treated that as settling ChatGPT's *client* surface. It doesn't.

`developers.openai.com/codex/mcp` now 308-redirects to `learn.chatgpt.com/docs/extend/mcp`, which says verbatim:

> "The ChatGPT desktop app, Codex CLI, and IDE extension support MCP servers and share MCP configuration for the same Codex host."

and documents the desktop UI path:

> "Open **Settings**, then select **MCP servers**. Select **Add server**. Enter a name, choose **STDIO** or **Streamable HTTP**, and provide the server's command or URL. Save the server, then select **Restart**."

MCP config lives in `~/.codex/config.toml` — the same file `--client codex` already writes.

So **ChatGPT Desktop can host a local stdio MCP server**, and this PR's hard refusal ("there is no local/stdio connector option to choose") is itself false — the exact class of misdirection the PR exists to remove. Only **ChatGPT web** is remote-only ("ChatGPT web doesn't read local Codex configuration files").

Fixing: `chatgpt-desktop` will install by writing the shared Codex TOML config rather than refusing. Also flagging that `docs/specs/ai-client-compatibility.md` §Phase-0 item 2 rests on the same false premise ("All ChatGPT surfaces are remote-only") and needs correcting before it merges.

PR #316, opened 2026-07-12
Make MoneyBin installable from PyPI: wheel packaging, release pipeline, httpfs seal

bsaffel:
### Why this exists

Publishing to PyPI today would ship a package that **cannot create a profile or run a transform.** The built wheel shipped none of the SQL schema, migrations, SQLMesh models, or synthetic demo data the runtime loads — the `[tool.setuptools.package-data]` globs pointed outside the package directory (which setuptools silently ignores), and the SQLMesh project was located by walking *up* out of the package (`parents[2]`), which only resolves in a source checkout. Half the wheel's runtime resources were missing and nothing caught it, because the test suite always runs inside the repo where the source tree fills the gaps.

This branch fixes the packaging, proves it with a test that inspects a real built wheel plus a CI job that installs the wheel into a clean venv **outside the repo** and runs the real user path, then adds the release pipeline to publish it safely.

### What's here

**Packaging correctness (the spine):**
- **`moneybin --version`** — a gap for any distributed CLI, and a dependency of the smoke test and the pinned config.
- **The SQLMesh project moved into the package** (`src/moneybin/sqlmesh/`) — one package-relative resolution rule that works identically in dev and installed. Exposed as the public `SQLMESH_ROOT` constant so tests import it instead of each re-deriving the path.
- **The wheel now ships every runtime resource** (150/150, was 75/150) — including three Python SQLMesh models and 28 Python migrations a naive glob list would miss. A hermetic, deny-list-based test inspects a real built wheel and fails if any on-disk resource is absent, so this cannot regress silently.
- **Clean-install smoke test** (`scripts/smoke_install.sh`, shared by the PR guard and the release matrix) — builds the wheel, installs it into a venv with no repo on `sys.path`, and runs `--version` → `demo` → `system doctor` from outside the checkout.

**Distribution:**
- **`moneybin mcp install`** now emits a **version-pinned** published-package config (`uv tool run --from moneybin==X.Y.Z`) when run outside a repo checkout, instead of pointing at the author's local tree. Pinning is deliberate: MoneyBin runs forward-only migrations when it opens a database, so an unpinned config would let a new release silently migrate a user's encrypted ledger on their next client restart. Upgrades stay explicit — re-run `moneybin mcp install`. The dev path (`uv run --directory <repo>`) is unchanged inside a checkout.
- **Tag-triggered release pipeline** (`release.yml`): guard (tag ↔ version ↔ CHANGELOG agree, and the tag is on `main`) → re-verify the tagged commit with the full CI and Scenario suites (made reusable via `workflow_call`) → build once → `pip-audit` → clean-install smoke on `{ubuntu, macos} × {3.12, 3.13}` → **human approval on a protected environment** → PyPI Trusted Publishing (OIDC, no stored token, PEP 740 attestations) → post-publish install-from-the-real-index check. The `make audit` target is now the single home for the pip-audit invocation, shared by `security.yml` so the accepted-vuln ignore list can't drift.

**Security fix (present on `main` today; surfaced while running the wheel like a user):**
- DuckDB disabled mbedtls encrypted *writes* in 1.4.1 (an RNG vulnerability), so the only encrypted-write path is now OpenSSL crypto — which ships inside the `httpfs` extension. Every encrypted database write was therefore silently auto-loading `httpfs`, leaving an **unrestricted remote http/s3 filesystem on the connection the MCP agent runs SQL against.** This branch makes the load explicit and seals the surface: it disables every remote filesystem `httpfs` registers (HTTP, S3, HuggingFace) and locks the configuration on the read-only connection — the only surface that runs agent free-text SQL — so an agent cannot re-enable remote access, re-install extensions, or unlock the config. An enumeration test asserts every filesystem `httpfs` registers is disabled, so a future DuckDB adding a fourth fails CI rather than silently opening egress.
- A misdiagnosed `duckdb<1.5.4` pin (an earlier theory for the encrypted-write failure) was reverted: 1.5.3 has the identical `httpfs` dependency, so the pin ceilinged a critical-path dependency for no benefit. `soupsieve` was bumped to clear two CVEs unrelated to this work that `pip-audit` flags on `main` too.

### Two things only you can do before the first tag

These are the human gate on an irreversible publish — the release workflow's approval step is a no-op until they exist:

1. **Create the PyPI + TestPyPI pending publishers** — project `moneybin`, owner `bsaffel`, repository `moneybin`, workflow `release.yml`, environment `pypi` (and `testpypi` for rehearsal).
2. **Create the `pypi` and `testpypi` GitHub Environments, each with yourself as a required reviewer** — that reviewer requirement *is* the human approval gate. Also add a repo ruleset restricting who may push `v*` tags, so the gate isn't the only thing between a stray tag and a permanently burned version number.

Recommend a `workflow_dispatch` → TestPyPI dry-run first: it exercises the environment gate, the OIDC config, and the publish job's permissions without burning a real PyPI version.

### Follow-up (tracked)

The `httpfs` extension is still fetched from `extensions.duckdb.org` on first run — a supply-chain and offline-first gap. Vendoring, pinning, and checksum-verifying it is its own spec.

### Not done here

The README presents the published `uvx moneybin` path as **arriving with the first tagged release** rather than claiming it works today — the package isn't on PyPI until the steps above are done and a release is tagged. The `git clone` path remains the working install until then.


bsaffel:

PR #318, opened 2026-07-12
Add Plaid Investments sync (M1G.4)

bsaffel:
## Summary

Brings Plaid's investment data into MoneyBin: securities, investment transactions, and dated holdings snapshots (with per-lot tax data) ride the existing `sync pull` job into five new `raw.plaid_*` tables and flow into the investment ledger. The shipped cost-basis engine derives lots, realized gains, and holdings with no engine changes.

The hard problem is **security identity**. Two Plaid items can describe the same instrument differently, and a wrong merge fuses two securities' tax lots — corrupting cost basis in a way the user cannot see and cannot easily undo. So the resolver refuses to guess: it adopts an existing binding, auto-binds only on an unambiguous strong identifier (CUSIP, ISIN, or exact ticker), and on *any* ambiguity mints a provisional security and files a merge decision for human review.

## Background

Two decisions are worth the reviewer's attention.

**Refuse-to-merge is absolute, and the danger isn't ties — it's false uniqueness.** A tie is safe: the ladder sees two candidates and refuses. What's dangerous is a lossy transform that manufactures a *unique* match where none exists, routing around the refusal entirely. Ticker suffix-stripping (`HEI.A` → `HEI`) and Unicode case-folding both do this. Neither is used for binding; only exact matches bind.

**Provider holdings are a claim, not the truth.** Broker snapshots land as non-authoritative `provider_reported_*` columns; MoneyBin derives holdings from its own lots and reconciles the two in `system doctor`. This is what lets the doctor catch a phantom position rather than inherit it.

## Changes

- **Ingestion** — five raw tables, loader extensions, UTC normalization of provider instants, and the orphan guard extended to the investment arrays.
- **Security identity** — the adopt-or-mint ladder (`SecurityResolver`), `app.security_links` / `app.security_link_decisions`, migration V035 (adds `created_by` to `app.securities`), and an audited merge service whose whole cascade undoes atomically.
- **Modeling** — Plaid staging, `core.fct_investment_transactions`, `core.dim_holdings`, the opening-lot bootstrap that seeds pre-window positions, and three new SQLMesh audits.
- **Surfaces** — `investments securities links {pending,set,history}` on CLI and MCP, plus eight investment reconciliation checks in `system doctor`.
- **Changes to existing behavior** *(the parts that touch code already on `main` — review these closely)*: the three pre-existing SQLMesh audits were missing `standalone TRUE` and had **never executed**, so the doctor had been reporting zero SQLMesh invariants since they shipped; `accounts_links_set` / `merchants_links_set` now require a human confirm to accept a merge and take an explicit action rather than inferring reject from an empty target; and `repoint` on the merchant and security link repos emitted its audit rows in the wrong order, so undoing the undo of any accepted merge failed deterministically.

## Test plan

- [ ] `make check test` — 5362 passed, 4 skipped (pre-existing)
- [ ] `make test-scenarios` — 77 passed
- [ ] Verify the resolver refuses to auto-bind on a stripped ticker, a Unicode-folded ticker, or an identifier tie
- [ ] Verify an agent cannot accept a link merge over MCP without a human confirm, and that a client which cannot elicit is directed to the CLI rather than proceeding
- [ ] Verify a fully-liquidated brokerage account is reported as a phantom holding rather than silently retaining its last-reported positions

## Breaking changes

- Migration **V035** adds `created_by` to `app.securities`; it applies automatically on the next database open.
- Five new `raw.plaid_*` tables are created by `init_schemas` — no manual step.
- **`accounts_links_set` and `merchants_links_set` change shape.** They now take an explicit action, and accepting a merge requires a human confirm. Any agent caller that relied on an empty target id meaning "reject" must now pass the reject action explicitly — an empty target is an input error.
- **The client-visible MCP tool surface goes from 102 to 105** (the three `investments_securities_links_*` tools). Stating it explicitly because the count is pinned against the live registry as a public contract: Cascade (Windsurf) holds at most 100 active tools across *all* of a user's MCP servers, so this widens an existing overflow from 2 to 5. That is the headroom tax that already demoted Windsurf to a documented-only tier rather than a release-gated one — no change to that call, but the guide, the compatibility spec, and the `mcp install` warning now cite the real number.

## Known limitations

Three behaviors ship a conservative default pending Plaid Sandbox golden validation, each marked `-- GOLDEN-GATED` in the SQL: reinvest/corporate-action pairing (`event_group_id`) is not yet linked, fee inclusion in `amount` is assumed (with a drift guard that warns if the assumption breaks), and every stock split routes to manual review instead of auto-deriving a multiplier.

## Related

Follows #311, #312. Refs #305.


bsaffel:
@claude the last couple reviews you've done didn't post any results.  are we good to approve?

bsaffel:
Fixed nine of the eleven 🟡 items in `4a393ed6`. Pushing back on one, and one is subsumed. Each was verified against the code before I touched it — details where the fix diverged from what was suggested:

**#8 (refresh gate)** — real, and worse than described. I counted the resolver's *actual writes*, not its outcome labels: `adopted` is the steady state and usually writes nothing, so counting outcomes would have fired a full transform run on every cash-only pull for the life of the database. Tests cover both directions (self-heal refreshes; a no-op re-resolve does not).

**#9 (timeout race)** — real. Root cause is that the per-call holder was a `threading.local`, which does not survive `asyncio.to_thread`'s thread hop, so an async tool body's write never registered and could commit after the caller got a `timed_out` envelope. Fixed by making the holder a `ContextVar` (which `asyncio.to_thread` propagates) and setting it around both branches — one change covers all three link tools rather than per-tool wiring.

**#11 (item drift)** — real. Worth noting it caught a bad fixture: `plaid_investments_sync_response.yaml` had one `account_id` reported under two items, which Plaid's item-scoped account ids make impossible. The Alpha-side position now sits on the Alpha account.

**#3 (ref conflict)** — real defect, but filing it into the review queue would make it worse: `SecurityLinksService.accept_merge` refuses to merge away a user-authored security, so a filed decision could be one the user cannot action, and routing a merge of two *established* securities through a path built for provisional→survivor risks exactly the tax-lot fusion the resolver exists to prevent. Surfaced through `system doctor` instead (`investment_conflicting_security_refs`), which is where a conflict with no safe automatic remedy belongs. The misleading log line — it pointed at a queue that never received the conflict — now points at the doctor.

**#2 (window-start message)** — subsumed by #11. The "item absent from metadata entirely" case can no longer reach the window check: a holding can only carry an item its own account belongs to, and the drift guard rejects anything else with an error naming both sides. Adding a branch for it would be dead code.

**#4 (`exchange` / `seeds.exchange_mic_map`) — declining.** The code fact is right (no JOIN) but the conclusion is the opposite of the design. `sync-plaid-investments.md` states it explicitly: *"Plaid delivers the exchange already as an ISO-10383 MIC, so it needs no normalization here; the resolver normalizes the catalog side (free-text `app.securities.exchange`) to a MIC through the seeded registry before comparing"* — and the spec names the alternative (normalize-at-write) as a deliberate, documented divergence, listing "promoting exchange to a stored normalized MIC column" as a possible future refinement. No production model reads this view's `exchange` at all: `core.dim_securities` reads `app.securities` directly. Adding the JOIN would force normalization onto user-authored catalog rows, which is the thing the design chose against. Happy to reconsider as a follow-up if the stored-MIC column becomes worth having, but it is not a defect in this PR.

Gates: unit 5554, integration 254 (+1 expected xfail), scenarios 78, lint/types clean.

bsaffel:
@claude nothing posted again, are we good to approve?

PR #328, opened 2026-07-17
Multi-currency capture & schema (M1K.1 Part A)

bsaffel:
## Summary

Implements M1K.1 Part A of `docs/specs/multi-currency.md` (Requirements 1, 2, 3, 8): captures original currency at every transaction/balance ingestion grain, stops the pipeline from silently defaulting unknown currency to `USD`, and gives transactions/balances an account-currency inheritance fallback. No conversion, no home-currency setting, no no-blend guard — those are a separate follow-up plan (M1K.1 Part B).

- **`dim_accounts.iso_currency_code` → `currency_code`** — direct rename (no deprecation shim, pre-launch), including the `accounts_set` MCP parameter. The CLI's `--currency` flag name is unaffected.
- **OFX `<CURDEF>` and Plaid balance `iso_currency_code`** now captured end-to-end (raw → staging) instead of being discarded.
- **`prep.int_transactions__unioned` no longer defaults unknown currency to `USD`** on any of its four arms (OFX/manual/tabular/Plaid) — this was the live correctness bug the initiative exists to close. Closing it required two follow-up fixes beyond the original scope: three upstream layers that made the manual-entry arm's fix a no-op (service-layer INSERT, a raw-schema `DEFAULT 'USD'`, a staging `COALESCE`), and the CLI's `transactions create --currency` option, which itself defaulted to the literal string `"USD"`.
- **Account-currency inheritance** added to `core.fct_transactions.currency_code` and `core.fct_balances`/`core.fct_balances_daily.currency_code`: a row's own captured currency wins, else the owning account's explicit `currency_code` setting. An account with no explicit setting still falls back to `USD` in `dim_accounts` (unchanged from before this PR) — a true "genuinely unknown" terminal case is scoped to M1K.1 Part B's no-blend guard, not this PR.
- **Regression fix:** the full whole-branch test run (not exercised by any single task's own tests) surfaced two real scenario failures — `core.dim_accounts`'s currency becoming genuinely nullable broke `matching/transfer.py`'s `a.currency_code = b.currency_code` blocking predicate (`NULL = NULL` is `NULL`, not `TRUE`, so nearly every transfer pair was silently excluded once currency stopped being hardcoded to USD). Fixed with a NULL-tolerant predicate that still blocks pairs with different *known* currencies.
- An end-to-end test drives a EUR-denominated OFX statement through the real `ImportService` import path (not raw-table seeding) confirming currency survives unmangled into both `core.fct_transactions` and `core.fct_balances`.

Docs updated: `multi-currency.md` spec promoted `ready` → `in-progress`, `INDEX.md`, CHANGELOG, roadmap, and `features.md`.

## Test plan

- [x] `make check` (ruff, pyright) clean
- [x] `make test-all` (unit + integration + e2e): 6382 passed, 4 skipped, 1 xfailed
- [x] `make test-scenarios`: 79 passed, 0 failed
- [x] `/code-review medium --fix`: 8 finder angles, correctness/reuse/conventions clean; one documentation fix applied
- [x] Iterative automated review (Codex + Claude Review) — every 🔴/🟡 finding verified against code and fixed or explicitly deferred to M1K.1 Part B with rationale


bsaffel:
@codex review

Fixed the 🟡 CONSIDER: renamed the remaining `iso_currency_code` references in `account-management.md`, `moneybin-mcp.md`, `account-identifiers.md`, and `data-model.md` to `currency_code` (commit 76299a19). `data-model.md`'s currency-handling note also needed a behavioral update, not just a rename — `fct_transactions.currency_code` now inherits from the account before falling back, it doesn't just default to USD.

Left the 🔵 NIT (`raw_manual_transactions.sql`'s dropped `DEFAULT 'USD'` with no companion migration) as-is — already reviewed twice during implementation and confirmed inert (the sole writer, `transaction_service.py`'s `create_manual_batch`, always passes `currency_code` explicitly, so the stored default never fires regardless of DB vintage).

bsaffel:
@codex review

Addressed all three findings from the last review pass:

1. **🔴 `unofficial_currency_code` missing for Plaid balances** — fixed (commit 4795ad7e): `SyncBalance` now carries both `iso_currency_code`/`unofficial_currency_code` mirroring `SyncSecurity`/`SyncInvestmentTransaction`/`SyncHolding` exactly, migration V038 amended (still unmerged, safe to amend in place) to add the column, staging passes both through, and `core.fct_balances`'s plaid CTE now does `COALESCE(iso_currency_code, unofficial_currency_code)`. New integration test drives the real extractor `.load()` + SQLMesh pipeline with only `unofficial_currency_code` set and asserts the currency survives to `core.fct_balances`.

2. **🟡 V036 rename breaks undo for pre-migration `account_settings.set` audit events** — fixed (commit b4c23b00): V036 now also rewrites historical `app.audit_log.before_value`/`after_value` JSON payloads for `target_table='account_settings'`, renaming the `iso_currency_code` key to `currency_code` while preserving the value and every other key. Scoped, NULL-safe (INSERT/DELETE-type events), idempotent, structural (dict-key) rewrite — not a string substitution, so it can't touch a payload whose *value* happens to equal that string. New tests cover the rewrite, cross-table scoping (including a payload that actually contains the old key, so the scoping predicate is genuinely exercised), and both NULL sides.

3. **🟡 `data-model.md` missing `currency_code` on `fct_balances`/`fct_balances_daily`** — fixed (commit 3101a75e).

All three fixes went through their own implement→adversarial-review cycle before landing, including one round where your reviewer's own follow-up caught that my first pass at the negative-case test for #2 didn't actually exercise the scoping predicate — fixed in eca3138a.

Full suite green: `make check`, `make test-all` (6381 passed), `make test-scenarios` (79 passed).

bsaffel:
@codex review

Fixed the currency-format-validation gap (commit 5ae0b1aa): promoted the check to a shared `moneybin.services._validators.validate_currency_code` helper (already the home for `validate_note_text`/`validate_slug`), refactored `AccountSettings.__post_init__` to use it (pure extraction, behavior-preserving), and wired it into `TransactionService._validate_manual_entry`. MCP `transactions_create` inherits it for free since it routes through the same `create_manual_batch` → `_validate_manual_entry` path.

Self-reviewed and caught one gap before you would have: the delegated error message dropped the `entries[{idx}]` attribution every sibling check in that function carries — fixed in 6eacb003, with the test tightened to assert the prefixed message.

Full suite green: `make check`, 6382 unit/integration/e2e, 79 scenarios.

bsaffel:
@codex review

Fixed `transaction-curation.md`'s stale `DEFAULT 'USD'` schema doc and `--currency` default text (commit a27f76bd) — both left over from before this PR's manual-arm currency fix. Confirmed `docs/specs/investments-data-model.md`'s `DEFAULT 'USD'` hits are the (unrelated, out-of-scope) investments domain, not stale.

bsaffel:
@codex review

Fixed both findings from the latest review pass (commit c6b8c6a2):

1. **🟡 `_validators.py:37` regex uses `.match()` not `.fullmatch()`** — fixed. `_CURRENCY_RE.match(value)` against the `^[A-Z]{3}$` pattern let a trailing-newline value (`"USD\n"`) through, since Python's `$` matches immediately before a final `\n`. Switched to `.fullmatch()`, matching the sibling `_SLUG_RE.fullmatch()` convention a few lines up. Added a dedicated `test_validators.py::test_validate_currency_code_rejects` parametrization covering this exact case plus the standard shape violations, alongside an `_accepts` case.

2. **🟡 Missing dedicated migration tests for V037/V038** — fixed. Added `tests/moneybin/test_migration_v037.py` and `test_migration_v038.py`, following the `test_migration_v030.py` pure-additive-DDL precedent: reconstruct the pre-migration table shape, insert a realistic row, run `migrate()`, assert the new column(s) exist and the pre-existing row gets `NULL` (no backfill), and assert idempotency on a second run.

Also noted and accepted your deferral on `dim_accounts.sql`'s `COALESCE(..., 'USD')` — that's the documented, tested M1K.1 Part B scope boundary (`test_transaction_currency_falls_back_to_account_default_when_neither_known`), not a gap this PR introduced. I'll tighten the PR description's "else NULL" wording per your note before this merges.

Full suite green: `make check`, 37/37 on the touched test files plus the full account/transaction service suites.

PR #329, opened 2026-07-17
Extract real credit-card PDFs (wrapped header, sections, MM/DD year)

bsaffel:
## Problem

Real Chase credit-card statements routed to an opaque seed
(`no_transaction_table`) even though the PDF is full of extractable text.
Three characteristics of real card statements — none reproduced by the
synthetic PDF fixtures — defeated recipe derivation, so the suite stayed
green while every real statement seeded:

- **Wrapped column header.** The header spans two physical lines
  (`Date of` / `Transaction … $ Amount`), so no single line splits into a
  date-led ≥3-cell header and the name-based reconstructor finds nothing.
- **Interleaved section sub-headers.** `PAYMENTS AND OTHER CREDITS` /
  `PURCHASE` / `INTEREST CHARGED` sit between the rows and broke the
  contiguous-run collector.
- **Year-less MM/DD dates.** Rows print `MM/DD` with no year; the year lives
  only on the `Opening/Closing Date` line, and the billing cycle crosses
  year-end (Dec → Jan).

## Fix

- **`auto_derive`** — reconstruct the transaction table from **data-row
  shape** (date-led, money-tailed) when no line names the columns, guarded
  by a "no named header" check so debit/credit statements still defer to the
  bridge.
- **`recipe` executor** — support the year-less `%m/%d` format and bracket
  each row's year from the statement's billing period, self-extracted from
  the document text so replay stays correct month to month.
- **`metadata`** — add `Opening/Closing Date` period anchors and
  `Previous/New Balance` anchors, with 2-digit-year parsing.

The card sign convention (`negative_is_income`) and its confirm gate are
unchanged — this only makes derivation reach them.

## Verification

- Verified against real statements (both card products, several months):
  every statement derives, extracts all rows, and **reconciles to ±1¢**
  against the stated opening/closing balances.
- New `chase_card_unruled` reportlab fixture reproduces the real shape
  (wrapped header + sections + MM/DD), plus:
  - two `auto_derive` unit tests (shape reconstruction; year inference across
    the year boundary),
  - a scenario test importing the fixture end-to-end (year-inferred dates,
    card-flipped stored amounts, reconciliation).
- Full unit suite + PDF scenario suite green; `make check` clean.


bsaffel:
@claude — CI is fully green (14/14) and this round is **0 🔴 MUST FIX**. The 4 remaining 🟡 CONSIDER are being **explicitly deferred** — invoking your approval contract's "fixed OR explicitly deferred" clause — tracked as a follow-up for the F10 live re-verify against real Chase statements:

1. `auto_derive` leap-day (`02/29`) *derivation-time* format detection — `_all_parse` trips the implicit non-leap-1900 default, though the executor's `_resolve_yearless_date` already handles leap days at cast time.
2. `_has_shape_transaction_rows` unbounded single-line match — one incidental date/money summary line classifies a non-transaction PDF as `transaction_table_underivable`.
3. `_parse_date` fixed-format whitelist ignores a bridge recipe's declared `date_format` for a custom period anchor.
4. The **replay** branch of the routing `YearlessDateError` handler has no `route_pdf_import`-level test (the first-contact branch is covered by `test_route_yearless_drift_row_fails_extraction_to_bridge`).

**Rationale for deferral:** these are shape-bounding / coverage residuals for statement shapes the shipped Chase fixture doesn't exercise. Fixing them correctly needs real-statement data (each blind attempt this round introduced a new edge case), which the owed F10 live re-verify provides. Critically, **all four fail-safe to seed today — no silent corruption** in the shipped path; the core derivation logic you've repeatedly confirmed correct is unaffected.

Given explicit, tracked deferral of these four, are we good to approve? If any one is a genuine blocker that can't be deferred, say which and why.

PR #357, opened 2026-07-25
Add M3K.3 spec for CLI output coherence

bsaffel:
## Summary

Adds `docs/specs/cli-output-coherence.md` (M3K.3), the design contract for making the CLI's text output coherent. A 2026-07-25 audit ran every representative command against a populated profile and read what came back: three commands that return rows each render in a different idiom, four print MCP tool names or service function names at the user, `reports spending` elides every value at 80 columns, and `moneybin refresh` narrates one pipeline stage of three. This spec defines one way to render each kind of result so those stop being independent judgement calls.

Spec only — status `draft`, no code changes. Scoped to the CLI's **text** branch; the JSON envelope, redaction, and privacy-audit paths are untouched, and command taxonomy stays with `moneybin-cli.md`.

## Changes

- **The render contract** — three named renderers (`render_rows` / `render_summary` / `render_note`) become the only way a command prints, plus per-report `DEFAULT_COLUMNS` with `--wide`, and a single `format_money` whose `+`/`−` is redundant with colour so the sign survives a pipe, a non-TTY, and `NO_COLOR`. This is the section worth reviewing first.
- **Requirement 35 is deliberately restrictive** — no renderer may deduplicate, merge, or suppress rows. `reports networth` has a separately-tracked model defect that makes an account appear once per balance source; collapsing those rows in the renderer would make the output *look* correct while the total stayed wrong. Worth a look to confirm the constraint is drawn in the right place.
- **Message hygiene is test-enforced, not advisory** — requirements 16/17/32 are grep-shaped audits over rendered output, and the tests enumerate commands from the live registry, so a new command cannot silently skip the contract.
- **Addressing** — filed as M3K.3 under the existing "CLI / MCP UX standards" milestone (after `.1` agent-visualization, `.2` mcp-tool-surface-scaling), with the matching `INDEX.md` row and 📐 roadmap entry per `shipping.md`.

## Test plan

- [ ] `pre-commit run --files docs/specs/cli-output-coherence.md docs/specs/INDEX.md docs/roadmap.md` passes — **run, green**
- [ ] All 8 relative links in the spec resolve from `docs/specs/` — **verified**
- [ ] Requirements are numbered 1–35 with no gaps, and all 10 internal cross-references resolve — **verified by script**
- [ ] No `private/` path appears in any of the three files — **verified**; `documentation.md` forbids public docs linking there
- [ ] Reviewer: confirm M3K.3 is the correct address, and that the `INDEX.md` row lands under the right section
- [ ] Reviewer: confirm requirement 9's "fits in 100 columns" threshold is the right bar for `DEFAULT_COLUMNS`
- [ ] Reviewer: sanity-check that requirement 35 doesn't over-constrain a legitimate future grouping/rollup renderer


bsaffel:
Requirement 31 needs a carve-out before implementation starts — as written it would hide a working command.

> 31. Unimplemented commands are hidden from `--help` output at every level (`hidden=True`) …

The obvious implementation is `grep -rn "_not_implemented" src/moneybin/cli/` → add `hidden=True` at each site. That sweeps in `review`, which is **partially** implemented, not stubbed: it prints queue counts, and `--confirm`/`--reject`/`--confirm-all` decide matches end-to-end. Hiding it removes a working command from `--help` — the opposite of the trust repair F4 exists for, and it would delete the only discoverable path to five review queues.

The real stub inventory is 9 whole commands across 3 modules — `sync key rotate`, `sync schedule {set,show,remove}`, `budget {set,delete}`, `transactions categorize ml {…}` — plus one partial in `review`. The count reduces the requirement to a short enumeration rather than a grep.

Suggested wording, appended to 31:

> **Scope is whole-command stubs only.** A command that reaches `_not_implemented` on *some* paths is not a stub and is not hidden. The enumeration is the four `sync` stubs, the two `budget` stubs, and the three `transactions categorize ml` stubs — not a grep for `_not_implemented`.

And in **Testing**, the bullet `- No stub command appears in any --help output (31)` needs its complement, or it passes trivially by hiding everything:

> - No stub command appears in any `--help` output, and every partially-implemented command still does (31).

Per `testing.md` ("a fixture that trips two guards isolates neither"), the one-sided assertion is the one that lets the over-broad implementation through.

Context: #358 just landed the `review` side of this — counts are now the default and the stub moved behind an explicit `--interactive`, so the only reachable `_not_implemented` in that command is one a user asked for by name. That makes the carve-out cheaper to state but no less necessary: the `_not_implemented` call is still there for a grep to find.

PR #361, opened 2026-07-26
Sub-total each currency instead of blending them (M1K.1)

bsaffel:
## Summary

Closes M1K.1 (Requirements 4–7) of [`multi-currency.md`](docs/specs/multi-currency.md): a profile can declare its home currency, and every report that sums money sub-totals each currency instead of adding dollars to euros.

The live defect: `reports.*` aggregated `amount` across currencies with no conversion, so a mixed-currency profile got a number denominated in nothing. Nothing flagged it — the blend was silent, and `dim_accounts` guessed `'USD'` for any account without an explicit currency, so even "unknown" looked like dollars. Both halves are fixed here.

**A single-currency profile — the common case — sees the same figures it always did**, plus the currency they are denominated in. That is held by fixtures, not by assertion: no scenario `.expectations` file changed, and the pre-existing report and net-worth tests assert the same numbers unchanged.

## Background

The one-way door this phase commits to is that **original currency is the canonical stored amount at every grain; conversion happens at presentation time, never by mutating stored originals**. Segmenting now is what makes that safe to build on: M1K.2 adds conversion on top of rows that already know what they are denominated in, rather than retrofitting a unit onto figures that lost it.

Two rules fall out and are enforced here:

- A `NULL` currency is **its own segment**, never resolved to the home currency. That guess would be invisible to every downstream guard.
- `home_currency` is `app.*` state, not `config.yaml`, because the SQLMesh models that read it cannot read YAML. It is written through a `*Repo` with paired audit emission (Invariant 10).

## Changes

- **Segmentation (Reqs 5, 7).** All seven `reports.*` models carry `currency_code` and group by it. `balance_drift` projects it without regrouping — asserted and computed balances belong to the same account, so it is single-currency by construction. `large_transactions` needed more than a `GROUP BY`: its median/MAD baselines and top-100 rank were pooled across currencies, scoring a typical dollar charge as an anomaly against euro-scale amounts.
- **Consumers that re-aggregate.** Segmenting a view is not enough if something downstream regroups it. The `core:cashflow` runner now carries `currency_code` in `group_cols` unconditionally for every `by` value, and `NetworthService` — whose `LIMIT 1` returned one arbitrary currency's total as the whole position, and whose history `LAST(net_worth ORDER BY balance_date)` silently picked one currency — partitions both.
- **Ranking and labelling.** Four runners ranked segmented rows with one cross-currency `ORDER BY … LIMIT`; at the default `top=25` an entire currency could fall out of the result. `top` on `core:merchants` and `core:large_transactions` now means top-N *per currency*. The `reports networth` CLI printed the alphabetically-first account's subtotal as the whole position; it now prints one headline per currency. `summary.display_currency` follows the rows instead of always claiming `USD`, and the `sql_schema` examples no longer hand agents queries that re-blend.
- **Account currency provenance (Req 3 completion).** `dim_accounts` resolved `COALESCE(s.currency_code, 'USD')` from a user-set-only column, and `fct_transactions`/`fct_balances` both COALESCE onto that — so one literal was the terminal fallback for the whole ledger and no row could reach core with a `NULL` currency. That made the unknown-currency segment, the doctor `fail` branch, and Req 8 unreachable in production. An account now takes the currency its own source reported (OFX `CURDEF`, Plaid `iso_currency_code`, the tabular `currency` column), resolving user override → source-declared → `NULL`. OFX and Plaid imports are unaffected — those formats always state a currency.
- **Home currency + doctor (Reqs 4, 6).** `app.profile_settings` (singleton, V044) behind `ProfileSettingsRepo`; `profile set home_currency EUR` and two MCP tools. `system doctor` gains `currency_integrity`: **fail** on unknown-currency rows with the `accounts set --currency` fix attached, **warn** on two or more known currencies, **pass** otherwise, publishing `moneybin_profile_currencies` and `moneybin_unknown_currency_rows{grain}`.

## Test plan

- [x] `make check test` — 7,642 passed, 4 skipped, 0 pyright errors
- [x] `make test-scenarios` — 83 passed; `tests/integration/` green
- [x] Every new guard written test-first and watched fail; each mutation-checked by restoring the defect (removing `currency_code` from `cash_flow`'s `GROUP BY` fails the scenario; the pre-fix cross-currency `ORDER BY` drops the USD row from the top-25)
- [x] `/code-review xhigh` — 15 findings, 14 fixed, 1 deferred (MCP eval fixture, tracked)
- [x] Codex review round — 3 findings, all 3 fixed. The unknown-currency scenario was mutation-checked against the restored `'USD'` literal and fails for the right reason.
- [ ] **Reviewers:** confirm the per-currency `top` semantics are the right call — it changes what `top=25` means for a mixed-currency profile (invisible for single-currency)
- [ ] **Reviewers:** confirm nulling the `core:networth` headline on mixed currency beats returning one currency's figure

## Breaking changes

Pre-launch, so no migration path is provided; these are contract changes teammates should know about.

- Every money-summing `reports.*` view gained a `currency_code` column, and each is now one row **per grain per currency**. A consumer that re-aggregates one of these views must keep `currency_code` in its own `GROUP BY` or it re-blends.
- `NetWorthSnapshotPayload` gained `currency_code` and `per_currency`; its headline scalars are `null` when more than one currency contributes. `NetWorthHistoryPoint` gained `currency_code`.
- `top` on `core:merchants` and `core:large_transactions` is now per-currency.
- `summary.display_currency` on report responses names the currency the rows are actually in; `ReportSemantics.currency` points at `currency_code`.
- `core.dim_accounts.currency_code` is `NULL` when no source stated a currency, where it was previously `'USD'`. A CSV without a currency column now reports unknown rather than dollars, and `system doctor` fails until it is set.
- `summary.display_currency` is nullable. A response whose rows span more than one currency — or whose currency is unknown — reports `null` rather than naming one of them.
- The MCP `profile` payload has no `database_exists` field — it could only ever be `true`, since the tool reads the database to answer at all.


bsaffel:
Round 7 response — three fixed, one declined with rationale, three were already fixed in the commit under review.

**Fixed**

- `schema_catalog.py` — the second `reports.cash_flow` example grouped by `category` alone and summed `outflow` across currencies. `sql_schema` serves that to agents as the recommended pattern, so it taught the exact blend this branch closes. Now grouped and ordered by `currency_code`.
- `docs/reference/data-model.md` — the `'USD'` fallback sentence contradicted both the model and its own next sentence.
- `summary.display_currency` scoping — see below.

**Declined, with the reasoning recorded in the tree**

`net_worth.sql`'s NULL pooling. The finding is accurate: `GROUP BY currency_code` puts every unknown-currency row in one segment and sums it, so two accounts in genuinely different currencies that both lack a stated one are added together. It is not fixable by grouping — nothing distinguishes two unknowns.

The alternative is withholding the segment's aggregates. That was rejected deliberately: the population that newly has unknown currencies is tabular imports whose file carried no currency column, who are almost always single-currency, so the pooled figure is the right one for them. Blanking net worth and cash flow for all of them trades a certain harm to the common case for a rare one that `system doctor` already reports as a **failure** — not a warning — with `accounts set --currency` as the named remedy.

Rationale and the rejected alternative are recorded in `docs/specs/multi-currency.md` (Requirement 5), the `net_worth.sql` column comment, and `data-model.md`, and it is tracked to be revisited alongside M1K.2, where an unconvertible segment needs an answer regardless. If it is ever taken, it must land across every money-summing model at once.

**On `display_currency` and the fetch boundary** — I did not withhold it on truncation. The field is documented as the currency of amounts *in the response*, and the resolver reads `records`, which is what the cursor fetched: `max_rows + 1`, one row more than is returned. Agreement across it is therefore always true of the returned rows and never a guess about rows past the boundary. Withholding whenever a result truncates would cost every large single-currency report a correct label and buy no safety, since the page-scoped claim was never wrong.

What was genuinely missing is that nothing *said* the claim is page-scoped, which is what allowed the over-read. The envelope docstring, the field comment, and the FastMCP session instructions now state that it describes the returned rows and that later pages may differ while `has_more` is true, and the truncation test asserts that semantic directly.

**Already fixed in the commit under review** (verified with `git show`, not the comments): `.claude/rules/mcp.md:155` was narrowed to profile *lifecycle* in `04fd1be9`; the `fct_balances` grain of `currency_integrity` has a test at `test_doctor_service.py:2079`, added in `307f34b3`; and the capability map carries a non-exempt `profile.settings` row covering both tools — `moneybin-capabilities.md` is a summary of that fixture, not a row enumeration, and its counts reconcile at 48 non-exempt rows over 49 tools.

bsaffel:
Round 9 — all three addressed at `fad5b1a0`. The 🔴 was correct and I confirmed it independently before touching anything.

**1. 🔴 `fct_balances_daily` blended transaction currencies into the carry — fixed.**

Traced it fresh rather than taking the report on faith. `txns` aggregated `SUM(amount) GROUP BY account_id, transaction_date` (no currency dimension), and `carry + txn_adj` added that to a balance whose `carry_currency` came from the observation. Since `fct_transactions.currency_code` resolves per row, a USD account holding a EUR charge summed unlike units.

The part worth naming is *why nothing caught it*: the blended row still reports the observation's currency, so `balance_drift`'s `currency_mismatch` check — added in this very branch — compares `USD` against `USD`, agrees, and passes the blend straight through to `reports.net_worth`. A guard downstream of an arithmetic blend can only compare labels, and both labels survive the blend intact. I recorded that as a placement rule in the spec, because it generalises past this one model.

The fix groups the aggregate by `currency_code` and applies only the rows denominated in the currency being carried. The excluded movement isn't dropped silently: it lands in the next observation's `reconciliation_delta` and reads as drift, which is what that field is for.

Test-first, and the fixture carries **both** a USD and a EUR transaction on the same day — with only the EUR one, a carry that correctly excludes it and a carry that never applied any adjustment produce the same number, so the test wouldn't have distinguished the fix from a no-op. Red at `-2.00` (= `873 − 875`, the blended carry), green at `-27.00`.

**2. 🟡 `currency_integrity` didn't mention the transform — fixed, and the neighbouring gap with it.**

Added the `moneybin transform` step to the fail detail, matching `dedup_reconciliation`'s convention in the same file.

I did **not** add a new invariant for the cross-currency case. The existing `len(currencies) > 1` warn already fires for it — its currency set unions `fct_transactions`, `dim_accounts`, and `fct_balances` — and its text claimed only that *reports* sub-total per currency. That was true before this branch and incomplete after it, since the carry now excludes foreign rows too. Amending the one message that was already firing beats standing a second invariant next to it.

**3. 🔵 `data-model.md` `dim_accounts` row — fixed.** It promised a `'USD'` default this branch removed, contradicting the corrected prose ~100 lines above. Not really a nit — the branch made that line false. Also documented the new carry contract in the `fct_balances_daily` section.

**Gates:** 7,655 unit (+2), 83 scenarios, full E2E + integration tiers, `pyright` clean. Scenarios and E2E were run because this changes a core model's arithmetic, not just its labels.

bsaffel:
Round 10 — the 🔴 was correct and is fixed at `39c79eeb`. Confirmed both call sites before changing anything: `account_service.py:445` (`str(row[6])`) and `:499` (`str(r["currency_code"])`), against `_load_settings` at `:385` which already passes `row[6]` raw.

The scope argument holds. `account_service.py` isn't in this diff, but removing `dim_accounts`' `'USD'` COALESCE is what makes `NULL` reachable in normal operation, and `currency_integrity` exists precisely to fail on those rows. My branch created the reachable state; the fix belongs here.

Worth naming: this file already contains a `TestNullAccountType` class whose docstring describes *this exact defect* — "The read path coerced the column with `str()`, which turns SQL NULL into the four-character string `"None"`" — fixed for `account_type` and left in place for `currency_code`, one line apart in the same constructor. The earlier fix treated the instance rather than the shape.

`AccountSummary.currency_code` and `AccountDetail.currency_code` are now `str | None`, and both `str()` wrappers are gone.

**One thing the finding didn't mention, which had to be fixed first.** Written as suggested, the new test failed with `'USD'`, not `'None'` — a wrong-reason failure. `tests/moneybin/db_helpers.py`'s `CORE_DIM_ACCOUNTS_DDL` still declared:

```sql
currency_code VARCHAR DEFAULT 'USD',
```

So the test double kept the default this branch removed from the model, and a fixture omitting the column came back `'USD'`. No test could reach the NULL state production now produces — the assertion would have passed against the bug for the wrong reason had I not checked the failure message.

`test_db_helpers_parity.py` exists to keep that DDL honest, but it compares column *name* sets, so a diverging `DEFAULT` is structurally invisible to it. Dropped the default; the full suite is green with it gone.

The new tests deliberately populate `account_type` and leave only `currency_code` unset — a row NULL in both would pass off the existing `account_type` fix and prove nothing about currency.

**Gates:** 7,657 unit (+2), 83 scenarios, full E2E + integration, `pyright` clean on both changed source files. I also checked every consumer of the two payloads for string formatting that would push the same rendering one layer out; `service_reports.py` and the investments CLI read different, already-nullable payload types.

bsaffel:

bsaffel:
## Round 11 — 1 🔴 + 3 🟡 addressed

All four verified against current code before acting; two of them turned out to have twins the review didn't name.

### 🔴 `profile show` crashes on a pre-V044 database — **fixed**

Confirmed exactly as described: `Database.__init__` returns at the read-only branch (`database.py:538`) before `init_schemas` and the migration runner, so `app.profile_settings` is absent on any database that hasn't had a write-mode open since V044. `classify_user_error` doesn't recognise `CatalogException`, so `handle_cli_errors` re-raises it.

**The MCP twin has the identical shape** — `mcp/tools/profile.py:38` also does `get_database(read_only=True)` → `ProfileSettingsService(db).get_settings()`. The review named only the CLI path, so the fix went into `ProfileSettingsRepo.get_home_currency()` where both inherit it, following the existing `CatalogException` guard in `account_link_decisions_repo` / `security_link_decisions_repo` rather than adding a second pattern.

Three tests, each confirmed to fail against the restored bug: repo-level, `profile show` (asserts `exit_code == 0` and `result.exception is None`), and the `profile` MCP tool.

### 🟡 `reconciliation_delta` blends across a currency change — **fixed**

Verified: with two observations in different currencies the delta was `Decimal('-100.00')` — a EUR-labelled USD difference, invisible to `currency_mismatch` exactly as described. It's now `NULL` when the observed currency differs from the carried one, which `balance_drift` already renders as `no-data`.

Recorded in `multi-currency.md` beside the existing guard-placement rule, since this is a second instance of it.

### 🟡 `schema_catalog` examples rank across currencies — **fixed, plus a derived guard**

Rather than patch the two named examples, this adds `test_every_currency_segmented_report_example_names_its_currency`, which derives the rule from each report's own declared `ReportSemantics.currency` — no hand-maintained list, so a new currency-segmented report inherits the guard.

It immediately caught a **third** offender the review didn't list: `reports.balance_drift`'s example ordered by `drift_abs` across currencies. Also fixed the `core.fct_balances` / `core.fct_balances_daily` examples, which the guard can't reach (not `@report` views) but which have the same defect.

### 🟡 balance payloads carry no currency — **fixed in-PR**

Judged in-scope rather than deferred, because this PR ships the server-instructions text making the claim the surface contradicts ("each row's own `currency_code` is then the authority"), and `core.fct_balances_daily` gained `currency_code` *in this PR* — so the data was already there.

- `BalanceObservationRow` gains a nullable `currency_code`; all three `BalanceService` reads project it.
- `display_currency` is now resolved from the rows instead of defaulting to `"USD"`, via `resolve_display_currency` promoted to `protocol/envelope.py` — the reports framework's `_resolve_display_currency` now delegates to it, so both surfaces answer identically rather than having two rules.
- **The CLI twin had it too**: `accounts balance show/history/reconcile` had the same hardcoded USD envelope. Fixed, and the text output now prints the currency beside the amount.
- Swept the rest of the surface for the same defect: of 25 money-bearing payloads, one other envelope never set `display_currency` (`accounts_get`, whose `credit_limit` is money). Fixed too, so the sweep is complete rather than partial.

While fixing this I found `test_balance_service.py` kept its own copy of the `fct_balances_daily` DDL — its docstring said "when that model adds a column, this CREATE must follow", and this PR's `currency_code` had not followed. It now uses the shared `db_helpers` DDL, which is the copy `test_db_helpers_parity` actually guards.

### Deferred (tracked)

`core.uncategorized_queue` has no `currency_code` column at all, so its example ranks unlike units by `priority_score` and the new guard can't see it (it isn't a `@report` view). Adding a column to a `core.*` model this PR never touched is wider than the finding, and fixing the column without fixing the currency-blind ranking would look addressed while still comparing a 50,000 JPY charge against a 500 USD one. Tracked with both parts together.

bsaffel:
## Round 12 — 1 🔴 + 3 🟡 addressed, plus the Codex P1 thread

Both reviewers converged on the same defect class, and checking it properly showed **my round-11 disposition was wrong**. I claimed there of having swept 25 money-bearing payloads and found exactly one gap. This round I enumerated the live registry instead of reading call sites:

**11 money-bearing tools, 9 of which never named a currency.** Between you, you'd found three of them.

### The root cause is the default, not the call sites

`build_envelope`'s `display_currency` parameter defaulted to `"USD"`, and 235 of its 251 call sites inherit that for free. Rounds 10 and 11 both patched named call sites; the class reopened both times, because the next money tool still inherits the default.

So this round fixes the default. `build_envelope` now reads `currency_code` off the payload it was handed — the record itself, or the rows of its single primary list (auxiliary `warnings`/`errors` lists excluded, matching `_count_primary_lists`) — and applies the existing `resolve_display_currency`: one agreed known code, else null. An explicit argument still wins, through an `UNSET` sentinel so an explicit `None` stays a real answer — the reports framework resolves across every matching row rather than the returned page, and must keep that.

`SummaryMeta.display_currency`'s own field default was `"USD"` too, one layer down; that's now `None` as well. A second hardcoded `"USD"` default sat on `accounts.py`'s `_coarse_envelope` helper — also fixed.

This closes all four of the findings below in one place, plus six tools neither review named.

### 🔴 `accounts_set` echoes credit_limit as USD — **fixed**

Confirmed. Fixed by derivation rather than the one-line `display_currency=payload.currency_code`, so `accounts_get`'s sibling fix stops being a thing anyone has to remember to copy.

### 🟡 `accounts()` list envelope — **fixed** (same mechanism)

### 🟡 Codex P1: balance assertions and `accounts(view="list")` — **fixed**

`BalanceAssertionRow` carried no currency at all: `app.balance_assertions` stores none, so nothing on the response said what unit the number was in. An assertion is a statement about one account, so the row now carries that account's `currency_code`, **joined at read time** rather than stored — it cannot drift from `dim_accounts`. Nullable, because `dim_accounts.currency_code` genuinely is. The CLI prints it beside the amount.

### 🟡 `doctor_service.py` `affected_ids` — **fixed**

Now `account:` / `transaction:` prefixed, matching `orphan_app_state`'s convention in the same file.

### 🟡 The two coverage gaps — **one added, one was the wrong test**

`balance_drift`'s one-side-unknown branch: added, parametrized over all three unexercised combinations (observation-unknown, account-unknown, both-unknown).

The `fct_balances_daily` NULL→known transition: **that case is unreachable, and the obvious test asserts the wrong thing.** `dim_accounts.currency_code` is `ARG_MAX(currency_code, extracted_at) FILTER (WHERE NOT currency_code IS NULL)` — one value per account — and `fct_balances` projects `COALESCE(u.currency_code, a.currency_code)`. So an earlier observation stating no currency *inherits* the account's, and both rows reach the daily model already agreeing. The delta is then genuinely well-defined and voiding it would withhold a real reconciliation.

I wrote it to assert that, with the derivation spelled out so the next person doesn't re-derive the wrong expectation. The reachable unknown case — an account with no currency anywhere — is covered separately, and it reconciles normally.

**A correction while I was there.** I'd written that this test exercises `_to_currency`'s NaN normalisation. It doesn't. I removed that branch and all 14 balance tests still passed: DuckDB renders an all-NULL VARCHAR column as object dtype holding `None`, never NaN — only an all-NULL DECIMAL column comes back as float64 (so `_to_decimal`'s NaN branch *is* live). The guard is defensive; I've corrected both the test docstring and the production comment that overstated it, and kept the guard, since the failure it prevents is silent.

### The guard

With the default fixed, a new money tool whose payload carries a currency is correct automatically — nothing to forget. The residual risk is the other shape: a payload returning money with no currency field anywhere, which can only answer null.

`tests/moneybin/test_mcp/test_money_tools_name_their_currency.py` enumerates those from the live registry and pins both sets by **equality**, not count — a count fires when someone adds one and never when someone forgets. It also fails loudly rather than skipping if a tool's return type won't resolve, since a tool silently dropped from the enumeration is exactly the blind spot it exists to close. Verified it bites by deleting a `currency_code` field: it named the offending tool and the fix.

Seven tools legitimately have no currency to report (`transactions`, `reviews`, `import_files`, `import_preview`, `system_audit`, `investments_lots_select`, `transactions_categorize_rules`). They now say null instead of an unfounded `USD`. Threading currency through those payloads is a wider change than this PR and is tracked separately.

### Contract change

`summary.display_currency` flips from `"USD"` to `null` on responses that carry no money, and on money whose currency isn't recorded. A single-currency profile sees the same value it always did. Called out in the CHANGELOG and in `multi-currency.md`, which now records *why* the guard belongs at construction rather than at call sites — the same placement lesson as the reconciliation guard.

### Verification

`make check test` 7685 passed / 4 skipped · `pyright` 0 errors · `make test-scenarios` 83 passed · slow balance integration 14 passed. Test delta 7663 → 7685 matches the new non-slow tests exactly. Restoration-checked: putting the `"USD"` default back fails 12 of the new tests for the right reasons.

bsaffel:

bsaffel:

PR #371, opened 2026-07-28
Route PDF account identity through AccountResolver

bsaffel:
## Summary

PDF import built its account key as a string and never called `AccountResolver`, so a card arriving as both a PDF statement and an OFX file had nothing to be matched against and both halves loaded. This routes PDF through the same resolution ladder every other source uses, and fixes two defects found while doing so.

**Identity scope.** The `source_native` link was scoped to the per-file alias, while `_lookup_strong_ref` and the `stg_tabular__*` JOINs key on `(source_type, source_origin, ref_value)`. Every statement of one card missed its own prior link and minted a fresh account plus a fresh review candidate — monthly, even after a human accepted the previous merge. It is now scoped to the issuer, mirroring how the tabular path derives its origin. `raw.import_log` keeps the per-file alias, because revert uses it to name the `pdf_<alias>` view.

**Last-four capture.** `metadata.py` stopped at the first whitespace-delimited token. `Account Number: XXXX XXXX XXXX 1234` captured the bare mask, producing a digit-free account key — the origin of the `chase_xxxx` placeholder seen in a real ledger. Worse, an unmasked `1234 5678 9012 3456` captured the *leading* four and reported them as the last four, feeding a wrong value into the institution+last4 merge signal. The whole grouped run is captured and reduced to its trailing four, and masked tokens normalise so one card whose statements render `****1234` and `xxxx1234` no longer keys two ways.

Also: the `resolve()` call is guarded so a raise finalizes `import_log` as failed and emits the failure metric rather than stranding the row at `importing`; `ACCOUNT_LINK_OUTCOMES_TOTAL` is recorded as at every other call site; and the `account_id_override` branch that now reaches the resolver has coverage.

## Test plan

TDD throughout — each fix has a test that failed first for the right reason.

- `test_grouped_masked_account_number_keeps_the_last_four` — failed with `account_id='XXXX'`
- `test_grouped_plain_account_number_last_four_is_the_final_group` — failed with `'1234' == '3456'`
- `test_consecutive_statements_of_one_card_share_one_account` — failed with `2 == 1`, two canonical accounts for one card
- `test_to_account_number_mask_covers_every_branch` — two existing rows asserted the non-normalised behaviour that made one card key two ways; those expectations were corrected and grouped cases added
- `test_account_id_override_pins_identity_through_the_resolver` — new coverage for a previously untested branch

`make check` clean (0 pyright errors), 8177 passed / 4 skipped, 85 scenarios passed. (Unit count rose from 7748 after merging `main`, which brought its own new tests.)

## Deliberately deferred

Tracked, not forgotten. Each needs a design decision rather than a fix, and bundling them would make this diff unreviewable.

- **Two cards at one issuer sharing a last four can merge silently.** The native key is unique only per issuer+last4. Strictly better than the pre-fix state, where every masked statement collapsed onto one key, but it is a silent merge and should not stay indefinitely.
- **No pre-load confirm gate on the PDF path.** `_import_tabular` gates via `_gate_account_proposals` before resolving; PDF resolves directly.
- **Resolver writes survive a failed ingestion.** `in_outer_txn=False` means `resolve()` commits independently of the raw rows it describes.
- **Saved recipes freeze the old metadata anchors**, so the capture fix does not reach formats derived before it; existing profiles need the recipe re-derived.
- **No scenario-level PDF fixture.** `tests/scenarios/` has none, which is why this class of defect survived a green suite.
- **A reissued card fragments into a second account** — a pre-existing, cross-source gap in the identity model, unrelated to this change.
- **`--account-id` does not register the statement's own native key.** The override branch sets `masked_acct = None`, so the only `source_native` link written is self-referential; a later import of the same card *without* the override derives `chase_1234`, finds no strong ref, and drops to the weak institution+last4 pass, so the pin does not carry forward. Raised by both reviewers. Deferred because the fix is not mechanical: `account_id` is hashed into `transaction_id` (`content_key = ...|{account_id}`), so deriving the native key in the override branch rotates `transaction_id` for every transaction previously imported with `--account-id` — the same double-import hazard this PR already ships an upgrade note for, applied to a second population. It needs its own upgrade note and a scenario run. Not silent corruption in the meantime: the weak-signal review queue still catches the second statement.


bsaffel:
@claude Re-review requested.

**No code change since your last pass.** Head is still `f109aebe`, CI is green (19 passing, 0 failed), security clear. What changed is the PR description.

Your remaining 🟡 — `account_id_override` not registering the statement's own native key — is **deferred, not fixed**, and it is now listed explicitly in the PR description's "Deliberately deferred" section. Your objection was that it was *"not fixed **or explicitly deferred**"*; the second half is what changed.

**Why deferred rather than fixed.** The suggested shape (derive `masked_acct` even when overriding, use the override only for `explicit_account_id`) is not mechanical, because `account_id` is hashed into `transaction_id`:

```python
content_key = (
    f"{period_marker}|{date_iso}|{raw_amount}|{raw_debit}|"
    f"{raw_credit}|{desc}|{account_id}"
)
raw_hash = content_key if dup_idx == 0 else f"{content_key}|{dup_idx}"
digest = hashlib.sha256(raw_hash.encode()).hexdigest()[:16]
transaction_id = f"pdf_{digest}"
```

Deriving the native key in the override branch therefore rotates `transaction_id` for every transaction previously imported with `--account-id`. That is the same double-import hazard this PR already ships an upgrade note for (`raw.tabular_transactions` won't collide on `on_conflict="ignore"` — the new rows land *alongside* the old ones), applied to a second population of users. It needs its own upgrade note and a `make test-scenarios` run, which is a materially wider blast radius than the rest of this diff.

It is not silent corruption in the meantime: the weak-signal review queue still catches the second statement, as you noted.

This is the same disposition you accepted for the other deferrals on the previous round — *"disclosed with clear rationale and tracked, not silently dropped."* Please re-review and approve if it meets that standard; if you think the `transaction_id` rotation is wrong or avoidable, say so and it goes back in scope.

bsaffel:
@claude Taking you up on "flag it and I'll re-evaluate."

**You're right that my rotation rationale was wrong**, under the reading you've now made explicit. If `account_id` stays pinned to the override and only `source_account_key` changes, no `transaction_id` moves. I had read the original suggestion as changing `account_id` itself. That part of the deferral rationale is withdrawn.

**But the two fields can't diverge, and here's the reason.** They're independent as *parameters* and coupled by the staging join.

`prep/stg_tabular__transactions.sql:73-79`:

```sql
LEFT JOIN app.account_links AS links
  ON links.status = 'accepted'
  AND links.ref_kind = 'source_native'
  AND links.source_type = ranked.source_type
  AND links.source_origin = ranked.source_origin
  AND links.ref_value = ranked.account_id      -- raw's account_id
```

`resolve()` writes exactly one `source_native` link per call, keyed on `source_account_key` — `account_resolver.py:380-390`:

```python
self._links.insert(
    link_id=uuid.uuid4().hex[:12],
    account_id=account_id,
    ref_kind="source_native",
    ref_value=src.source_account_key,   # the only ref_value written
    ...
)
```

and the explicit-binding branch (`:143-149`) routes straight through that same single write.

So the join requires a `source_native` link whose `ref_value` equals **raw's `account_id`**. Apply the fix as specified — `account_id` = override, `source_account_key` = `chase_1234` — and the only link written is `chase_1234 → my_pinned_account`. Nothing has `ref_value = 'my_pinned_account'`, the LEFT JOIN misses, and **every transaction in that import silently stops canonicalizing.** That trades a forward-teaching gap for a live canonicalization regression on the same rows.

The self-referential link isn't dead weight — it's what satisfies that join today. That's why it's there.

**The actual fix is both links**, not a swap: keep `my_pinned_account → my_pinned_account` so the join still resolves, and add `chase_1234 → my_pinned_account` so a later un-overridden statement hits the strong ref. `app.account_links` is many-to-one, so that's legal. But `resolve()` writes one link per call, so it needs either a second write or a resolver change — and `_write_native_mapping` raises when a `source_native` already maps to a different `account_id`, so the interaction with that guard needs deciding too.

That's more than decoupling two fields, and it's the same conclusion by a different route: a design decision, not a mechanical fix. Codex's P2 proposes the swap shape and hits the identical problem.

Happy to be checked on this — if the join tolerates the miss somewhere I haven't traced, say so. Otherwise I'd ask that it be accepted as a deferral on the corrected rationale, and I'll update the PR description to state this reason rather than the rotation one.

PR #439, opened 2026-08-23
Flag a source-native key that is really a canonical account id

bsaffel:
Adds the `app_account_links_canonical_native_key` doctor invariant.

A source-native key is what a *file* calls an account — a slugified label, a `pdf_doc_*` digest, a content key. A canonical `account_id` is what MoneyBin calls it. An `--account-id` pin used to overwrite the first with the second, binding a document under an id no source could ever derive.

The check reports accepted `source_native` links whose `ref_value` matches a `core.dim_accounts.account_id`.

**`warn`, not `fail`.** Such a link still resolves the document to the right account: the invariant is broken, but no row is wrong, and re-importing the file normalizes it.

**Stacked on #438.** That PR is what makes a canonical-shaped native key underivable. Without it this check reports on current behavior rather than guarding a contract — a pinned PDF import on `main` today produces exactly the shape it looks for. Merge #438 first; GitHub will retarget this to `main` automatically.

### It does not escalate on a "forked" origin

An earlier revision failed when one `source_origin` carried several native keys, on the theory that the document had been imported twice. That is wrong twice over:

- `source_origin` is the **exporter/format** identity for tabular sources (`matched_format.name` — "monarch", "tiller"), not a document. Every account in one multi-account export shares it.
- A pin used to *teach* the resolver the document's own key alongside the canonical one — two links, one import, nothing duplicated.

Link rows cannot distinguish either shape from a genuine double import; only the raw key-spaces can. The check no longer claims otherwise, and a test pins that.

### Verification

- 9711 passed, 4 skipped; 92 scenarios
- `ruff` clean, `pyright` 0 errors
- On a live profile: 1 affected account → `warn`, matching the hand-derived prediction

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,7 +18,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
   binding the document under an id no source could ever derive. The new
   `app_account_links_canonical_native_key` invariant `warn`s on those links:
   they still resolve to the right account, so the invariant is broken but no
-  row is wrong, and re-importing the file normalizes it. Now that pinned
+  row is wrong. Re-importing writes the document's own key alongside the old
+  link rather than retiring it — there is no automatic cleanup for these links
+  yet, and the warning says so instead of promising one. Now that pinned
   imports keep the source's own key, nothing can derive a canonical-shaped
   native key, so this is a regression guard on that contract.
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,17 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 ## [Unreleased]
 
 ### Added
+- **`doctor` catches a source key that is really a MoneyBin account id.** A
+  source-native key is what a *file* calls an account — a slugified label, a
+  `pdf_doc_*` digest, a content key. A canonical `account_id` is what MoneyBin
+  calls it. An `--account-id` pin used to overwrite the first with the second,
+  binding the document under an id no source could ever derive. The new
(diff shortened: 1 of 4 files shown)


bsaffel:
Closing this. The check cannot be made sound against the current schema, and I reproduced its core failure end-to-end:

```
moneybin import standard.csv --account-id primary-checking --account-name "Primary Checking"
  -> app.account_links row ("primary-checking", "primary-checking", "accepted")
  -> doctor: warn, affected_ids=["primary-checking"]
     "1 account(s) store a canonical account id as a source-native key
      (a legacy --account-id pin) ... no automatic cleanup exists yet."
```

Nothing legacy, nothing pre-fix, nothing wrong -- and the message tells the user it can never be cleared.

Root cause: step 0 of the resolution ladder (`account_resolver.py`) writes `explicit_account_id` **verbatim** as `account_id`. There is no `uuid4().hex[:12]` mint and no validation on that path, so "this value appears in `account_links.account_id`" proves nothing about provenance. Pre-fix residue and a perfectly ordinary pin produce byte-identical rows; the link table cannot distinguish them. Three successive predicates here failed for three distinct reasons (`core.dim_accounts` COALESCE fallback, then link-membership, then this), which is the signal that the premise is wrong rather than the query.

There is also a privacy edge: `affected_ids` carries the account source-native key, and for OFX that is the institution real `<ACCTID>`, reaching CLI JSON and MCP labelled as a safe internal id.

Nothing depends on this check. #438 fixes the underlying defect and already supersedes stale raw rows automatically on re-import. A sound version needs mint provenance recorded as an explicit fact at mint time -- a schema change, and its own piece of work rather than a follow-up here.

Thanks to both reviewers for staying on this one; the repeated flags were correct each time.

PR #519, opened 2026-09-04
Adopt moneybin.tables constants at executed-SQL call sites

bsaffel:
## Summary

The `moneybin.tables` `TableRef` registry existed as the canonical source of schema-qualified table names, but was bypassed by hardcoded string literals (e.g. `"app.audit_log"`) at roughly 90 executed-SQL call sites across the codebase — meaning a future table rename could silently miss a call site. `services/undo_service.py`, the one service whose job is restoring corrupted `app.*` state, had no `moneybin.tables` import at all and referenced `app.audit_log` nine times as a bare literal.

This closes three registry gaps (`meta.model_freshness`, `app.metrics`, `prep.stg_security_prices` had no constant) and sweeps the hardcoded literals to the matching `TableRef` constant. Pure mechanical adoption — every constant's `.full_name` renders to the exact string it replaced, so no query text changes and no behavior changes.

## Changes

- Added `MODEL_FRESHNESS`, `METRICS`, and `STG_SECURITY_PRICES` to `src/moneybin/tables.py`, following the existing shape.
- Swept ~90 literal sites to constants across 22 modules: services (`undo_service`, `audit_service`, `transaction_service`, `doctor_service`, `transform_service`, `system_service`, `demo_service`, `schema_catalog`, `import_service`), extractors (`plaid`, `tabular/formats`), `investments/sqlmesh_loader`, `matching/priority`, `metrics/persistence`, `migrations.py` (the runner, not the frozen historical scripts), `privacy/sql_lineage`, `repositories/securities_repo`, `connectors/gsheet/pull_service`, `synthetic/reset` and `synthetic/writer`, and two CLI commands.
- In the four files a parallel PR is restructuring import machinery in (`import_service.py`, `account_links_service.py`, `refresh.py`, `inbox_service.py`), touched only SQL string literals plus one added `moneybin.tables` import per file where needed — no import reordering, no `# noqa: PLC0415` lines touched. `account_links_service.py`, `refresh.py`, and `inbox_service.py` turned out to already be fully on constants; `import_service.py` had 8 real sites plus one deferred-import block extended in place.
- Deliberately left untouched: `src/moneybin/sql/migrations/*.py` (frozen, point-in-time DDL — must track the schema as it existed when it ran, not a live registry that can rename underneath it), user-facing hint text in error/detail messages that names a table for a suggested `moneybin db query` command, and `schema_catalog.py`'s `EXAMPLES` dict values (illustrative SQL text served to agents via the schema doc, never executed by MoneyBin itself).

## Test plan

- [x] `make check` (ruff format, ruff lint, pyright) — clean, 0 errors
- [x] `make test` — 10,544 passed, 0 failed, 4 skipped (pre-existing, unrelated)
- [x] `make test-integration` — 578 passed, 0 failed
- [x] `make test-scenarios` — 92 passed, 0 failed
- [x] Isolated adversarial pre-push review of the full base-to-head diff, independent of this session — verified every substitution renders byte-identical SQL, then flagged one real gap (`synthetic/writer.py` had five ingest-target literals in the same shape the Plaid-extractor sweep already converted) and a cosmetic noqa-comment inconsistency; both fixed in a follow-up commit, then the full three-suite gate re-run clean.

### DIFF
--- a/src/moneybin/cli/commands/stats.py
+++ b/src/moneybin/cli/commands/stats.py
@@ -13,6 +13,7 @@ import typer
 from moneybin.cli.output import OutputFormat, output_option, quiet_option
 from moneybin.cli.utils import handle_cli_errors
 from moneybin.database import get_database
+from moneybin.tables import METRICS
 from moneybin.utils.parsing import parse_duration
 
 logger = logging.getLogger(__name__)
@@ -78,7 +79,7 @@ def stats_command(
                                    PARTITION BY metric_name, metric_type, labels
                                    ORDER BY recorded_at DESC
                                ) AS rn
-                        FROM app.metrics
+                        FROM {METRICS.full_name}
                         {where_sql}
                     )
                     WHERE rn = 1
--- a/src/moneybin/cli/commands/synthetic.py
+++ b/src/moneybin/cli/commands/synthetic.py
@@ -63,13 +63,17 @@ def _run_generate(
             from moneybin.database import (
                 get_database,  # noqa: PLC0415 — deferred import
             )
+            from moneybin.tables import (
+                OFX_TRANSACTIONS,  # noqa: PLC0415 — deferred import
+                TABULAR_TRANSACTIONS,
+            )
 
             with get_database(read_only=False) as db:
                 # Check if profile already has data
                 try:
                     row = db.execute(
-                        """SELECT (SELECT COUNT(*) FROM raw.ofx_transactions)
-                                + (SELECT COUNT(*) FROM raw.tabular_transactions)"""
+                        f"""SELECT (SELECT COUNT(*) FROM {OFX_TRANSACTIONS.full_name})
+                                + (SELECT COUNT(*) FROM {TABULAR_TRANSACTIONS.full_name})"""  # noqa: S608  # TableRef constants
                     ).fetchone()
                     existing_count = row[0] if row else 0
                 except Exception:  # noqa: BLE001,S110 — tables may not exist in a fresh DB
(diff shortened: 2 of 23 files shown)


bsaffel:
## 🟡 CONSIDER — deferred to a tracked follow-up

Taking the deferral branch of the approval contract rather than fixing in-PR. Filed as **[MB-167](https://linear.app/moneybin/issue/MB-167/guard-against-reintroducing-hardcoded-schema-qualified-table-names-in)**, a child of MB-39, milestone R1.

**Why deferred rather than fixed here.** The guard has to statically decide *"is this string literal executed as SQL?"*, which is undecidable in general. A naive scan is either noisy — it flags every docstring — or vacuous, exempting so much that it catches nothing. That is the same shape of problem that took PR #512 sixteen review rounds, and it is the wrong thing to bolt onto a mechanical zero-behavior-change sweep whose whole value is that it is easy to verify.

**The follow-up is not a stub.** MB-167 carries the exemption taxonomy this PR already discovered, so the next author does not re-derive it:

- `src/moneybin/sql/migrations/*.py` — ~52 historical scripts, frozen point-in-time DDL that must *not* track a live registry. Zero of them import `moneybin.tables`; that is deliberate, not a gap. (`src/moneybin/migrations.py`, the runner, is a different file and is swept here.)
- Prose, docstring, and comment mentions — hundreds of hits.
- `services/schema_catalog.py`'s `EXAMPLES` values — illustrative SQL served to agents, never executed.
- User-facing hint and error text naming a table inside a suggested CLI command.
- `database.py`'s module-docstring usage example.

It also specifies the shape: follow `test_adapter_layering.py`'s allowlist convention, prefer a file-level allowlist plus scanning only literals that reach an execute call, assert exemptions by set equality so a stale exemption fails as loudly as a new violation, and mutation-prove each assertion.

**Scope note for the record:** the sweep landed 95 substitutions across 22 files, not the ~40 the ticket estimated.

bsaffel:
@claude The single 🟡 is deferred to MB-167 per the deferral branch of your approval contract — details in the comment above. Please re-review and confirm the verdict.

PR #558, opened 2026-09-07
Count bare list fields and check a declared row set names one

bsaffel:
## Summary

The sweep that makes every payload declare its row set is meant to fire *before*
the runtime raise in `moneybin.protocol.row_set`. For two shapes it never fires
at all, and the raise reaches a caller instead. Both are agreement failures
between the sweep's notion of "carries a collection" and `row_set_field`'s.

**A bare `list` annotation is invisible.** `typing.get_origin(list)` returns
`None` for an unparameterized annotation, so a helper keyed on `get_origin(hint)
is list` calls `rows: list` a scalar and drops the class out of the sweep.
`row_set_field` still raises, because it tests the *value* with `isinstance`.
No payload is annotated this way today and no lint rule forbids it.

**A declaration naming a scalar passes both gates.** `@row_set` checks at class
definition that the name is one of the payload's fields — never that it is a
list. A class carrying no list field is filtered out of the sweep, so nothing
fails until the first production call.

## Why the second case matters more than it reads

It is the shape a rebound decorator leaves behind. When a class inserted
directly above a decorated one steals its `@row_set`, the class that *lost* the
decorator is caught twice over — undeclared by the sweep, and raising at runtime.
The thief is the silent half: it keeps a declaration nobody wrote for it, and if
it happens to carry a field of the declared name, a diagnostic list becomes the
row set. That is a wrong `returned_count`, which is the defect this whole
contract exists to remove.

The issue proposed having `@row_set` record the qualified name it decorated and
assert the match. That cannot work: in a steal the decorator genuinely *does*
decorate the thief — Python binds it to whatever definition syntactically
follows — so the assertion compares a value against itself, and the inserted
class leaves no runtime trace. Intent is unrecoverable at runtime; only
coherence is. Decoration-time type checking is out for a separate reason: every
payload module uses `from __future__ import annotations`, so `get_type_hints`
fails on forward references at class-creation time. The sweep is the right home.

## What changed

- `_is_list_hint` counts bare `list`. It still does **not** count `tuple`:
  `row_set_field` decides with `isinstance(value, list)`, so counting tuples
  would demand a declaration for a class the runtime would never object to —
  a second disagreement in place of the one being fixed.
- A new sweep over payloads that *declare* a row set, asserting the declared
  name is one of the class's list fields. Deliberately not a subset of the
  classes with collections, since the escaping case has none.

## Test plan

Both halves were proved by mutation rather than by a green run.

- Injected both defect shapes as probe classes into a real payload module. Each
  guard fired, naming the class and the wrong field:
  `_MutationProbeBareList carries ['rows'] but declares no row set`;
  `_MutationProbeScalarDeclaration declares row set 'label', which is not one of
  its collection fields (none)`.
- Reverted **only** the `if hint is list:` line with both probes still in place.
  The bare-list probe vanished from the failure list, leaving just the scalar
  one — so that line causes the catch, and the two guards are independent rather
  than one masking the other.
- Collection reconciles rather than merely looking green: the file went 147 →
  253 collected, +106, which is the new 104-case parametrization plus the two
  standalone tests. `test_payload_declares_its_row_set` is unchanged at 146.
- `make check` — ruff clean, bare repo-wide `uv run pyright` 0 errors (3
  pre-existing `reportlab` stub warnings in a file this diff does not touch).
- `make test` — 11178 passed, 4 skipped. The 4 skips are pre-existing and
  environmental; none touch `row_set.py` or this test.

## Review evidence caveat

`chatgpt-codex-connector[bot]` reported at 2026-09-07T00:49:06Z that it has hit
its Codex usage limit and cannot review until credits are added. That is
account-wide, so an absent codex review on this PR is a non-verdict, not the
usual 👍-and-no-review-object clean signal. Please don't read silence from it as
a pass.

Closes MB-176.

### DIFF
--- a/tests/moneybin/test_protocol/test_row_set_declaration_coverage.py
+++ b/tests/moneybin/test_protocol/test_row_set_declaration_coverage.py
@@ -25,6 +25,7 @@ from __future__ import annotations
 
 import importlib
 import pkgutil
+import types
 import typing
 from dataclasses import fields, is_dataclass
 
@@ -50,9 +51,16 @@ def _is_list_hint(hint: object) -> bool:
     origin = typing.get_origin(hint)
     if origin is list:
         return True
-    if origin is None:
-        return False
-    return any(_is_list_hint(arg) for arg in typing.get_args(hint))
+    # Unwrap only what wraps a field's type without changing what the field
+    # holds. Recursing into any generic instead would find the ``list[str]``
+    # inside ``dict[str, list[str]]`` and call that field a collection, while
+    # ``row_set_field`` sees a dict and refuses it — the disagreement this
+    # helper exists to prevent, one container level down.
+    if origin is typing.Annotated:
+        return _is_list_hint(typing.get_args(hint)[0])
+    if origin is typing.Union or origin is types.UnionType:
+        return any(_is_list_hint(arg) for arg in typing.get_args(hint))
+    return False
 
 
 def _list_field_names(cls: type) -> list[str]:
@@ -229,3 +237,51 @@ def test_a_tuple_annotation_is_not_a_collection() -> None:
     """
     assert not _is_list_hint(tuple)
     assert not _is_list_hint(tuple[str, ...])
+
+
+@pytest.mark.unit
+def test_a_container_whose_values_are_lists_is_not_a_collection() -> None:
+    """A field holding a dict of lists holds a dict, and is not a row set.
+
+    ``get_origin(dict[str, list[str]])`` is ``dict`` — neither ``list`` nor
+    ``None`` — so unwrapping every generic rather than a named few reaches the
+    ``list[str]`` in its arguments and reports the field a collection.
+    ``row_set_field`` tests the value with ``isinstance(value, list)`` and
+    rejects the dict, so the sweep would demand a declaration the runtime would
+    never honour, and worse, would accept ``@row_set`` naming that field:
+    ``test_a_declared_row_set_names_a_collection`` asks whether the declared
+    name is in ``_list_field_names``, and the same over-count that put it there
+    is what makes the check pass. The mismatch would then surface only on the
+    first production call — the failure this module exists to move earlier.
+
+    No payload is shaped this way today, so this fixes no live miscount; it
+    holds the helper to the equivalence its docstring claims.
+    """
+    assert not _is_list_hint(dict[str, list[str]])
+    assert not _is_list_hint(tuple[list[str], ...])
+    assert not _is_list_hint(set[str])
+
+
+@pytest.mark.unit
+def test_the_wrappers_the_payloads_use_are_still_unwrapped() -> None:
+    """Every wrapper the sweep unwraps keeps a case here.
+
+    ``_is_list_hint`` names the wrappers it unwraps instead of recursing into
+    any generic, which makes dropping one a one-line edit that silently empties
+    part of the sweep rather than failing. ``Annotated`` is the case that would
+    hurt: every field in ``moneybin.privacy.payloads.system`` carries
+    ``Annotated[T, DataClass.X]``, so losing it would take
+    ``TransformPlanPayload`` and its four ``Annotated[list[str], ...]`` fields
+    out of the scan while the suite stayed green.
+    """
+    assert _is_list_hint(typing.Annotated[list[str], "metadata"])
+    assert _is_list_hint(typing.Annotated[list[str] | None, "metadata"])
+    assert not _is_list_hint(typing.Annotated[str, "metadata"])
+
+    # `X | None` evaluates to `types.UnionType`, so the legacy spelling is the
+    # only way to reach the `typing.Union` sentinel. Ruff's UP045 keeps it out
+    # of the tree, which makes that branch defensive rather than live — and
+    # exercising it here is what says so, instead of leaving a reader to guess
+    # whether it is dead.
+    legacy_optional = typing.Optional[list[str]]  # noqa: UP045  # branch under test
+    assert _is_list_hint(legacy_optional)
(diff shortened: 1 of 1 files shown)


bsaffel:
Both findings were right and both are fixed in `19156c53`.

## 🟡 CONSIDER — anti-vacuity on `_PAYLOADS_WITH_DECLARATIONS`

Fixed, and the consequence is worse than the review states, so it is worth recording precisely.

I proved it by mutation rather than by argument: pointing the filter at the wrong type (`isinstance(declared_row_set(cls), bytes)`) empties the list, standing in for the discovery break you describe. The run does **not** merely "report green while testing nothing" in the sense of passing 104 vacuous cases — all 104 disappear:

```
SKIPPED [1] tests/.../test_row_set_declaration_coverage.py:167: got empty parameter set for (payload_cls)
```

An empty parametrization is a **skip**, and a skip fails nothing. So the whole guard would have evaporated with the only trace being one skip line in a 297-test run — the exact silent-green failure this PR exists to close, reproduced one level up inside the guard itself. Restoring the filter and adding the assertion turns it into a named failure.

The check now mirrors the sibling: three classes pinned, one per scanned root — `AccountListPayload` (payload package), `ExportDestinationsOutput` (`moneybin.cli.output`), `ReportResult` (`moneybin.reports._framework.execute`) — plus a floor.

One deliberate difference from the sibling worth calling out. The sibling asserts `> 100` against a census of 146 (31% slack); the census here is 104, so copying `> 100` would leave 4% and red the suite on any legitimate payload removal. The floor is `75`, with the census and the reasoning in a comment. The pinned names are what actually make the check bite — the floor only catches a collapse.

I also extended the existing test rather than adding a second one. `_payload_classes()` is a single scan and both lists are filters of it, so "the scan reaches the payloads it is meant to guard" is one fact about one scan. Its docstring now says why the two need checking separately: one reads annotations, the other reads the declaration key, so they fail independently.

## 🔵 NIT — stale `pragma: no cover` reference

Fixed. You are right that it is drift I introduced in this diff — the previous commit renamed `_payload_classes_with_collections` to `_payload_classes` and left the comment pointing at the old name. Corrected to `_payload_classes`.

## Note on the hand-trace

Flagging for whoever reads this next: `uv` was not on `PATH` in the review sandbox, so `_is_list_hint` was verified by hand-tracing against stdlib `typing` rather than by running the suite. That trace matched the local run, so nothing is in dispute — but it means the review's evidence for the logic is analytical, not executed. The executed evidence is the mutation runs recorded in the PR body and above.

Gate after the fix: `make check` clean, bare repo-wide `uv run pyright` 0 errors, `tests/moneybin/test_protocol/` 297 passed / 0 skipped.

bsaffel:
Correct, and fixed.

`typing.get_origin(dict[str, list[str]])` is `dict` — neither `list` nor `None` — so the blanket `any(...)` branch reached the `list[str]` in its arguments and returned `True`, while `row_set_field`'s `isinstance(value, list)` refuses the dict. That is the disagreement this PR exists to close, one container level down.

Your reading of the knock-on is the part worth keeping. `test_a_declared_row_set_names_a_collection` asks whether the declared name is in `_list_field_names`, and the same over-count is what would have put it there — so the test I added in this PR to catch a misdirected declaration would have passed on exactly that declaration, and the `RowSetContractError` would have waited for the first production call.

Your grep matches mine: no payload has that shape, so nothing was miscounted in the tree. The censuses are unchanged — 146 collections, 104 declarations — and the file goes 253 → 255 tests, which is the two new cases and nothing else.

## The fix

The recursion now names what it unwraps instead of descending into any generic:

```python
if origin is typing.Annotated:
    return _is_list_hint(typing.get_args(hint)[0])
if origin is typing.Union or origin is types.UnionType:
    return any(_is_list_hint(arg) for arg in typing.get_args(hint))
return False
```

`Annotated` takes `get_args(hint)[0]` rather than `any(...)`: the remaining arguments are metadata, and a metadata object that happened to be a list type would otherwise vote.

## Two tests, because narrowing has its own failure mode

`test_a_container_whose_values_are_lists_is_not_a_collection` is the one you asked for — `dict[str, list[str]]`, `tuple[list[str], ...]`, `set[str]`. Restoring the blanket recursion fails it:

```
E   assert not True
     +  where True = _is_list_hint(dict[str, list[str]])
1 failed, 254 passed
```

The second test exists because your fix converts an over-permissive rule into a hand-maintained allowlist, and dropping a member of an allowlist is a silent narrowing. `test_the_wrappers_the_payloads_use_are_still_unwrapped` pins each member. Deleting just the `Annotated` branch:

```
10 failed, 213 passed
```

— the wrapper test, plus nine payloads (`TagsPayload`, `ImportStatusPayload`, `ReportCatalogPayload`, …) whose declared row set is an `Annotated[list[...]]` field, and the collection census falls 146 → 114. `moneybin.privacy.payloads.system` annotates every field as `Annotated[T, DataClass.X]`, so this is not a hypothetical branch: `TransformPlanPayload` carries four `Annotated[list[str], ...]` fields.

The tuple test stays as it was. It and the new one guard opposite directions and fail independently — `tuple[str, ...]` was already handled correctly by the old code, so it would not have caught this.

## One honest caveat on the `typing.Union` arm

It is defensive, not live. `X | None` evaluates to `types.UnionType`, so the only way to reach `typing.Union` is the legacy `Optional[...]` spelling — which ruff's UP045 rejects repo-wide, as it did to my first draft of the test. I kept the arm rather than dropping it, because its absence fails by silently shrinking the sweep rather than by raising, and suppressed the rule on the one line that exercises it so the branch is demonstrably live rather than left for a reader to guess at.

Gate: `make check` clean, bare repo-wide `uv run pyright` 0 errors, `tests/moneybin/test_protocol/` 299 passed / 0 skipped.

bsaffel:
@claude re-review requested — the automated run on `48bc347b` (workflow run 34073183480) completed successfully but posted no review object and no inline comment, so this PR currently has no verdict on its head commit. The last recorded verdict is `❌ NOT APPROVED` on `19156c53`, which predates the fix for that round's 🟡 CONSIDER.

State on `48bc347b`: the `_is_list_hint` over-count you raised is fixed (recursion now unwraps only `Annotated` / `Union` / `UnionType` rather than any generic), with two tests and mutation evidence recorded in the comment above. All 21 CI checks pass.

PR #563, opened 2026-09-07
Scope categorization coverage to the transactions needing a category

bsaffel:
Closes MB-155.

## The defect

Four surfaces answer "which transactions are uncategorized," and until now three of them disagreed:

| Surface | Population it counted |
|---|---|
| `core.uncategorized_queue` → `moneybin review` | `category IS NULL`, not a transfer leg, not an archived account |
| `categorization_coverage` doctor check | `category IS NULL`, not a transfer leg — archived accounts included |
| `categorize stats` / MCP `transactions_categorize_stats` | every transaction, against every row in `app.transaction_categories` |

So `categorize stats` could report a backlog of N while `moneybin review` handed the user a different number and then declared itself finished. MB-44 canonicalized the queue and deliberately left these two alone, because in each the word sits inside an arithmetic identity where swapping the numerator alone breaks the figure.

## The change

Both surfaces now read the population `core.uncategorized_queue` is drawn from, through **one** query rather than three hand-maintained predicates. `DoctorService._run_categorization_coverage` delegates to `CategorizationService.coverage()` instead of keeping its own SQL, so the two cannot report different coverage for the same database.

Three things fall out of scoping the denominator:

- **Transfer legs and archived accounts leave the count.** Neither is work anyone is asked to do, and `moneybin review` never offers them.
- **Categorizations orphaned from their transaction stop counting as coverage.** `categorized` was a row count over `app.transaction_categories`, which includes rows whose transaction no longer exists — doctor keeps a separate invariant for exactly those.
- **`by_source` gains a `source_supplied` bucket.** A transaction whose CSV column or manual entry carried category text of its own is categorized but has no assignment method. Without the bucket, the breakdown the CLI prints directly beneath `Categorized` would have stopped summing to it — the same reconciliation defect this PR fixes, one line lower.

## What users will notice

The numbers move, deliberately. A ledger whose transfer legs and closed accounts made up much of its "uncategorized" count now reports a smaller backlog and a higher percentage — the same data, described as work rather than as rows.

`total_transactions` in the MCP payload is now the size of that population rather than a count of all transactions. The field name is unchanged; its description, the payload docstring, and the CLI header all say what it now means (`Categorization coverage (excludes transfers, archived accounts):`).

## What deliberately did not move

`transactions list --uncategorized` still filters on `categorized_by IS NULL` — "MoneyBin never decided this one." It is the only surface that reveals categories a CSV or manual entry supplied but nobody confirmed; `moneybin review` skips those because they already carry text. Moving it would have made all four surfaces agree at the cost of the only window onto that population. Decided with Brandon rather than assumed, and the question underneath it — should the review queue be offering those labels for a quick keep/change at all? — is filed as MB-180.

## Test plan

Test-first; every assertion watched fail for its own reason before the implementation existed.

- `TestCoverageOverWhatNeedsCategorizing` (6 new unit tests) over a fixture holding every row shape the figures must classify exactly once — two bare rows, two categorized inside MoneyBin, one carrying source-supplied text, plus a transfer leg, an archived-account row, an unresolved-account row, and an orphaned categorization. Expected counts hand-derived before running (5 / 3 / 2, 60.0%).
- The anti-drift test builds `core.uncategorized_queue` from the **shipped model file** and asserts the stats figure equals its count, so a future divergence between the view's predicate and this query fails a test rather than quietly publishing a contradicted number.
- Two of the six assert the invariants that must survive the change (`total = categorized + uncategorized`; `by_source` sums to `categorized`). Both passed before and after — they are regression guards, not new behavior.
- `test_categorization_coverage_ignores_archived_accounts` for the doctor half. Its rows set `is_transfer` explicitly: left NULL they would be dropped by the transfer filter instead, and the test would have passed without exercising the archived-account one at all.
- Four pre-existing fixtures modeled databases that cannot exist — transactions with no `dim_accounts` row, categorizations with no transaction, an `app.transaction_categories` write not reflected on the fact table the production view COALESCEs it onto. Corrected rather than worked around; `_reflect_categorizations_into_fact` performs the join the view would have.

Gates run: `make check`, `tests/moneybin/test_services/test_categorization_service.py` (125 passed), `tests/moneybin/test_services/test_doctor_service.py` (125 passed), `make test`, `make test-integration`. `make generate-docs` produced no diff.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,7 +34,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
   column or manual entry carried category text of its own, so the breakdown
   still sums to `categorized`. `transactions list --uncategorized` deliberately
   does not move: it means "MoneyBin never decided this one", which is the only
-  way to find those source-supplied labels.
+  way to find those source-supplied labels. (#563)
 - **`moneybin refresh` says what each pipeline step did.** A run that changed
   nothing and a run that recategorized 400 transactions both printed a single
   `✅ Refresh complete in 4.2s`, because the counts each step computed went to
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,30 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 ## [Unreleased]
 
 ### Changed
+- **Categorization coverage counts the work you can actually do.** `moneybin
+  transactions categorize stats`, MCP `transactions_categorize_stats`, and the
+  `categorization_coverage` doctor check each derived their own "uncategorized"
+  figure, and none of them matched the one `moneybin review` hands you. Stats
+  counted every transaction in the ledger against every row in
+  `app.transaction_categories` — including transfer legs nobody is asked to
+  categorize, archived accounts nobody maintains, and categorizations orphaned
+  from transactions that no longer exist. Doctor counted a third population,
+  excluding transfer legs but not archived accounts. All of them now read the
+  population `core.uncategorized_queue` is drawn from, so the percentage
+  measures a backlog you can act on and every surface reports the same size for
+  it.
+
+  **The numbers move.** A ledger whose transfer legs and closed accounts made up
+  much of its "uncategorized" count will report a smaller backlog and a higher
+  percentage than before — the same data, described as work rather than as rows.
+  `total_transactions` in the MCP payload is now the size of that population
+  rather than a count of all transactions; the field name is unchanged, and its
+  description says so. `by_source` moves with it and gains a `source_supplied`
+  bucket for transactions that count as categorized only because their CSV
+  column or manual entry carried category text of its own, so the breakdown
+  still sums to `categorized`. `transactions list --uncategorized` deliberately
+  does not move: it means "MoneyBin never decided this one", which is the only
+  way to find those source-supplied labels.
 - **`moneybin refresh` says what each pipeline step did.** A run that changed
   nothing and a run that recategorized 400 transactions both printed a single
   `✅ Refresh complete in 4.2s`, because the counts each step computed went to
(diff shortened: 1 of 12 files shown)


bsaffel:
Both findings confirmed and fixed.

## 🟡 CONSIDER — `plaid_unmapped` scope overclaim

You're right, and the concrete shape you described is reachable: that query reads `prep.stg_plaid__transactions` directly with no `dim_accounts` join and no `is_transfer`/`archived` filter, so `total_transactions: 5` beside `plaid_unmapped: 400` is a valid response that my two new sentences told a reader was impossible. I introduced that claim in this PR; the field was already scoped differently before it.

**Took the carve-out rather than the scoping — not because it is cheaper, but because scoping would make the metric wrong.** `plaid_unmapped` answers "which Plaid category codes does my bridge not cover?" That is a property of the bridge, not of this user's backlog: whether a PFC code has a mapping does not change because one transaction carrying it happens to sit on an archived account or be half of a transfer. Narrowing it through `_CATEGORIZABLE_POPULATION` would under-report the very gap it exists to expose, and would also have to be re-grained onto the merged layer it deliberately avoids for cost.

Three changes:

- `CategorizeStatsPayload`'s docstring now names the five coverage fields explicitly instead of saying "every count", and gives `plaid_unmapped` its own paragraph: ledger-wide by design, may exceed `total_transactions`, not part of the reconciliation.
- The MCP tool docstring does the same in the terms an agent needs, ending with "Do not subtract or compare it against the coverage figures."
- A comment at the query itself in `queries.py` states that it is deliberately **not** scoped to `_CATEGORIZABLE_POPULATION`, and why. Without it the next reader reconciles the code to the docstring and silently narrows the diagnostic — the docstring alone would have made that look like the fix.

## 🔵 NIT — CLI header named two of three exclusions

Fixed rather than deferred; it cost one word. Now `Categorization coverage (excludes transfers, archived and unresolved accounts):`, which matches the MCP docstring and `docs/specs/moneybin-doctor.md`.

PR #564, opened 2026-09-07
Remove the inert Pylint suppressions and correct false cycle claims

bsaffel:
Closes MB-168.

`pyproject.toml`'s ruff `select` does not list `PL`, so every Pylint-family code
is disabled and no `# noqa: PLC0415` ever suppressed a diagnostic. They were
documentation comments wearing a lint-suppression costume — and because no
linter evaluated the justifications riding on them, wrong ones survived
unchallenged. An audit took one at face value and had to be retracted.

This removes all 363 of them, keeps the justifications that earned it, and
corrects the ones that were false.

## The claims were audited by experiment, not by reading

The ticket warns that a module-level import graph cannot show a cycle that a
working deferral is holding open — it says whether a cycle exists *now*, not
whether the deferral is load-bearing. So each of the 325 deferred imports in
`src/` was tested by actually hoisting it to module scope and importing the
module in a clean interpreter, in both directions (a cycle `A<->B` surfaces
only from whichever module the interpreter enters first).

The harness was calibrated first against the cycles the ticket names as
genuine, with the expected verdict written down before running. The first run
returned "no cycle" for all of them — which was the harness truncating
parenthesised multi-line imports at the comment, hoisting `from x import (`.
Without that calibration step the result would have read as "every claim is
false." Fixed, all five reproduce as real, and a malformed rewrite now reports
as a tool error rather than as a clean result.

## What the experiment found

|                      | Real cycle | No cycle |
| -------------------- | ---------- | -------- |
| **Claims a cycle**   | 8          | **17 false** |
| **Says nothing**     | **7**      | 293      |

- **17 false claims removed.** The ticket predicted 19 of 24; the measured
  figure is 17 of 25. Hoisting any of them raises nothing.
- **7 silent load-bearing deferrals now documented.** These are the ones the
  ticket did not know about: bare markers holding a real cycle open with no
  comment at all — `config.py -> moneybin.database` (x2),
  `transaction_categories_repo -> services.categorization._shared` (x3),
  `account_resolver -> import_service`, `import_inbox -> import_cmd`. Anyone
  hoisting one breaks the import; nothing said so.
- **8 true claims kept**, reworded to the verified fact.

One of the ticket's six named cycles, `errors.py -> services.profile_service`,
no longer exists — MB-51 removed it. `errors.py` has no reference to it.

## Why this is not `ruff --fix`

Ruff can autofix unused directives, but it parses this repo's dominant em-dash
form as part of the directive and deletes the justification with it:

| Before | After `ruff --fix` |
| --- | --- |
| `# noqa: PLC0415` | removed |
| `# noqa: PLC0415  # polars is not cold-start cheap` | comment kept |
| `# noqa: PLC0415 — defer import to keep CLI cold-start light` | **justification destroyed** |

About 80 of the 107 justifications use the em-dash form, so the autofix would
have silently erased them. Survivors are normalised to a plain `# reason`
comment, which is both what AGENTS.md asks for and immune to a future autofix.

## Disposition of all 363 markers

- **279 dropped** — 218 bare, plus notes that only restate what the reader can
  already see ("defer import" on an import sitting inside a function).
- **51 kept** — genuine non-obvious whys: cold-start cost, import weight,
  a `TYPE_CHECKING`-only module-scope import.
- **15 rewritten** to the verified cycle fact.
- **17 false cycle claims removed.**
- **1 preserved verbatim** — a `# pyright: ignore[reportPrivateUsage]` riding
  on the same line.

### The retained cold-start claims were measured too

The hoist experiment answers "does this cycle?", not "is this import expensive?",
so the kept justifications would otherwise have been carried forward on their
face — the same defect in a different coat. Each named module was imported in a
clean interpreter against a 16 ms bare-interpreter baseline:

- **13 of 20 distinct targets cost 85–260 ms** — `services.currency_service`
  260 ms ("polars is not cold-start cheap"), `privacy.payloads.imports` 232 ms,
  `services.import_service` 223 ms. Well founded.
- **Four cost under 15 ms** — `rich.text` 13 ms, `mcp.surface` 11 ms,
  `rich.cells` 7 ms, `connectors.sync_errors` 4 ms.

I left the weak four alone rather than deleting them. A standalone measurement
overstates independence: `rich.cells` is co-loaded with `rich.console` (23 ms)
anyway, so its marginal cost in situ is not what the number above says. The
measurement is strong enough to confirm a claim, not to refute one, and the
deferral itself is harmless either way. Flagging it rather than acting on it.

**Only comments change. No import is moved**, so the runtime import graph is
byte-for-byte what it was. That is checked rather than asserted: every one of
the 83 changed files parses to a byte-identical AST against the branch base.
Comments are not nodes, so an unchanged AST means nothing executable moved.

## Beyond the ticket

Two things worth flagging, neither acted on here:

1. **The count was ~6x the estimate** — the ticket says "roughly 60"; the
   actual figure is 354 `PLC0415` plus 9 other `PL` codes.
2. **1,164 further inert `noqa` markers remain** outside the `PL` family
   (`S608` x706, `BLE001` x190, `ARG001` x96, `SLF001` x67, ...), measured with
   `ruff check --extend-select RUF100`. Enabling `RUF100` permanently would
   close the whole class, but it cannot go green until those are swept. Worth
   its own ticket.

## The guard

`test_no_inert_pylint_suppressions.py` stops the markers re-accreting. It also
asserts its own premise — that `PL` is absent from ruff `select` — so if the
family is ever enabled the guard says to delete it rather than quietly becoming
the next false claim. It caught 9 non-`PLC0415` markers I had missed on the
first pass, which is why the sweep covers the whole `PL` family.

This guard is an addition beyond the literal ask; it is a two-way door and easy
to drop if you would rather not carry it.

No CHANGELOG entry: `.claude/rules/shipping.md` excludes internal changes with
no behaviour change.

## Test plan

- `make check` — 0 ruff errors, 0 pyright errors.
- `make test`, `make test-integration`, `make test-scenarios`, `make test-e2e`.
- The guard verified red against the pre-change tree before being relied on.
- `grep` confirms zero `# noqa: PL…` markers remain outside the guard's own
  docstring.

### DIFF
--- a/.claude/rules/cli.md
+++ b/.claude/rules/cli.md
@@ -95,9 +95,14 @@ Every E2E test, every shell autocomplete, and every CLI invocation pays the full
   ```python
   @app.command("serve")
   def serve(...) -> None:
-      from moneybin.mcp.server import build_server  # noqa: PLC0415 — defer import
+      from moneybin.mcp.server import build_server  # fastmcp is not cold-start cheap
       build_server(...).run()
   ```
+
+  Say *why* in a plain comment, and never reach for `# noqa: PLC0415`. Ruff's
+  `select` omits `PL`, so that marker suppresses nothing — it only looks
+  official, which is how 17 wrong justifications rode one unchallenged until
+  MB-168 tested them. `test_no_inert_pylint_suppression_markers` now rejects it.
 
 - **Verify with `importtime`.** When adding a new command module, confirm the cold-start path stays clean:
 
--- a/docs/reference/cli-startup-flow.md
+++ b/docs/reference/cli-startup-flow.md
@@ -75,12 +75,17 @@ Heavy transitive imports MUST be deferred inside the function that needs them:
 @app.command("list-prompts")
 def mcp_list_prompts(...) -> None:
     from moneybin.mcp.server import (
-        mcp,  # noqa: PLC0415 — defer fastmcp import to subcommand body
+        mcp,  # defer fastmcp import to subcommand body
     )
     ...
 ```
 
-The `# noqa: PLC0415` suppression is the grep-able marker for the pattern. The same pattern applies to anything that pulls a parser, ORM, or large package graph: `fastmcp`, `sqlmesh`, `polars`.
+A plain comment naming the cost is the whole convention — there is no
+suppression to add. Ruff's `select` omits `PL`, so a `# noqa: PLC0415` would
+suppress nothing, and `test_no_inert_pylint_suppression_markers` rejects it. The
+pattern itself is greppable by structure: an `import` indented inside a
+function body. The same applies to anything that pulls a parser, ORM, or large
+package graph: `fastmcp`, `sqlmesh`, `polars`.
 
 The non-obvious failure mode: `from x import Y` at module top of any command file (including transitive imports from helper modules those files load) loads the heavy graph for *every* invocation — `--help`, autocomplete, every E2E subprocess. The CLI feels fine in isolation but the test suite slows by a factor of N over time. The CI guard described below catches regressions early.
 
@@ -289,4 +294,4 @@ Before opening the PR, verify cold start stays clean:
 uv run pytest tests/moneybin/test_cli/test_cold_start.py tests/moneybin/test_cli/test_help_no_wizard.py -v
 ```
 
-If your new command needs a heavy import, put it inside the function body with `# noqa: PLC0415 — defer <dep> import` per the pattern in `mcp.py` and `transform.py`. See CONTRIBUTING.md → "Adding a new CLI command" for the full recipe (test layer expectations, JSON-output parity, `--help` audit).
+If your new command needs a heavy import, put it inside the function body with a plain comment naming the cost — `# <dep> is not cold-start cheap` — per the pattern in `mcp.py` and `transform.py`. See CONTRIBUTING.md → "Adding a new CLI command" for the full recipe (test layer expectations, JSON-output parity, `--help` audit).
--- a/src/moneybin/cli/commands/accounts/links.py
+++ b/src/moneybin/cli/commands/accounts/links.py
@@ -75,7 +75,7 @@ def links_pending(
     payload = AccountLinksPendingPayload.from_service(groups, n_pending)
 
     if output == OutputFormat.JSON:
-        from moneybin.cli.output import render_or_json  # noqa: PLC0415 — defer import
+        from moneybin.cli.output import render_or_json
 
         render_or_json(
             build_envelope(data=payload),
@@ -220,7 +220,7 @@ def _report_rematch(rematch: RefreshResult | None) -> None:
     # Deferred, and below the guard: matching_service pulls duckdb and the match
     # engine, this module is on the CLI cold-start path (.claude/rules/cli.md),
     # and the reject path reaches here needing none of it.
-    from moneybin.services.matching_service import (  # noqa: PLC0415
+    from moneybin.services.matching_service import (
         PENDING_MATCHES_HINT,
     )
 
@@ -316,10 +316,12 @@ def _plan_merge(
     """Resolve the merge against ``db``, the way MCP plans the same decision."""
     # Deferred: the identity contracts and decision service are not worth loading
     # on every CLI invocation to gate one subcommand.
-    from moneybin.protocol.write_contracts import (  # noqa: PLC0415 — keep off the cold-start path
+    from moneybin.protocol.write_contracts import (  # keep off the cold-start path
         AccountLinkDecisionRequest,
     )
-    from moneybin.services.review_decisions_service import (  # noqa: PLC0415 — keep off the cold-start path
+
+    # keep off the cold-start path
+    from moneybin.services.review_decisions_service import (
         ReviewDecisionsService,
     )
 
@@ -536,7 +538,7 @@ def links_history(
     payload = AccountLinksHistoryPayload.from_rows(rows)
 
     if output == OutputFormat.JSON:
-        from moneybin.cli.output import render_or_json  # noqa: PLC0415 — defer import
+        from moneybin.cli.output import render_or_json
 
         render_or_json(
             build_envelope(data=payload),
@@ -624,7 +626,7 @@ def links_run(
     )
 
     if output == OutputFormat.JSON:
-        from moneybin.cli.output import render_or_json  # noqa: PLC0415 — defer import
+        from moneybin.cli.output import render_or_json
 
         render_or_json(
             build_envelope(data=payload),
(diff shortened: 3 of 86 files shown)


bsaffel:
Round 1 addressed — all three findings were real. Head is now `eb2e7f1e`.

**🔴 MUST FIX — docs still taught the marker.** Correct, and it was a gap in my sweep: I scanned `*.py` and never grepped the prose that teaches the pattern, so an always-loaded rule file would have had the next contributor add a marker that fails the new guard on the following run.

- `.claude/rules/cli.md` — sample now uses the plain-comment form, plus a line saying why a Pylint `noqa` is never right here.
- `docs/reference/cli-startup-flow.md` — the quoted `mcp.py` snippet now matches its source again (this PR had edited that exact line), and the "the grep-able marker for the pattern" sentence is replaced rather than deleted: the pattern is greppable by structure — an import indented inside a function body.
- Confirmed no `PLC0415` reference survives anywhere outside the three places that now explicitly prohibit it.

**🟡 CONSIDER — redundant adjacent comments.** Dropped the generic line at the 9 sites where an existing docstring or comment already explained that cycle more specifically. Kept it at 6 where the neighbour covers something else — `orchestrator.py:178` and `:863` sit under comments about *gauge refresh*, not the cycle, and `config.py` ×2, `transaction_categories_repo.py:246` and `cli/output.py` had no existing explanation at all.

**P2 (Codex) — the premise test read only `select`.** The sharpest of the three, since it lands on the property I was most confident about. It now reads `select` and `extend-select` and treats `ALL` as enabling the family. Verified by mutation, not by reading: adding `extend-select = ["PLC0415"]` turns it red, so does `["ALL"]`, and it returns green on revert.

**Gates:** `make check` clean (0 ruff, 0 pyright). `make test` 11355 passed / 4 pre-existing skips — up from 11322 because the branch was rebased onto `9c1882f9`, which brought #562's tests in. `make test-integration` 592 passed. All 83 changed Python files remain AST-identical to the base, so this is still comments and prose only.

bsaffel:
Round 2 — the second Codex P2 was right, and sharper than the first. Head `c424593e`.

**The finding:** treating `select = ["ALL"]` as enabling the Pylint family ignores `ignore`. Ruff resolves the two by prefix specificity, so `select = ["ALL"]` with `ignore = ["PL"]` leaves the family **off** — my check would have failed CI instructing someone to delete a guard whose premise still held. The rule cuts both ways: `select = ["PLC0415"]` with `ignore = ["PL"]` leaves that one **on**.

**The fix is not to model that.** Re-implementing ruff's resolution here would repeat the mistake this PR exists to correct — reading configuration instead of measuring behaviour. `ruff check --show-settings` already reports `linter.rules.enabled` fully resolved, so the test asks ruff and looks for a `PL` code in the answer. That is immune to `select`, `extend-select`, `ignore`, specificity, and anything added later.

It also now fails loudly if the parse returns nothing, so "I could not read the rule set" can never be mistaken for "no Pylint rule is enabled" — that is precisely this guard's own failure mode, and the reason it exists.

**Verified by mutation, both directions:**

| `pyproject.toml` | Expected | Result |
| --- | --- | --- |
| `extend-select = ["ALL"]` + `ignore = ["PL", …]` | pass — premise holds | ✅ passes |
| `extend-select = ["ALL"]` alone | fail — premise broken | ✅ fails |

The previous version got the first row wrong, which is exactly what Codex flagged.

**Gates:** `make check` clean. `make test` 11355 passed / 4 pre-existing skips. `tests/moneybin/test_architecture/` 97 passed. The 83 Python files in the sweep are untouched by this round — the only change is the guard test, so they remain AST-identical to the base.

Note the one new suppression here is `# noqa: S404` on `import subprocess`, which is a **live** directive: `S` is enabled, so unlike everything this PR removes, it actually suppresses something.

bsaffel:
Round 3 — also a real gap. Head `c6168e45`.

**The finding:** the scan required the directive to open a comment, so ruff's file-level spelling (`ruff:` ahead of the directive) slipped past and left a route for an inert Pylint suppression to re-enter a scanned file. Not hypothetical — the repo already uses that spelling for `S101` and `A001`, so it is the form a contributor would reach for.

Widened to accept both spellings. **Verified by planting one**: a file-level Pylint directive at the top of `src/moneybin/cli/utils.py` turns the scan red (`src/moneybin/cli/utils.py:1`), and it goes green on revert. No file-level Pylint directive exists today, so the wider pattern flags nothing already present.

**One boundary stated explicitly** rather than left to look like an oversight — I checked the remaining spelling before it gets raised. Ruff treats the flake8-compatible file-level form as a *blanket* suppression and discards the codes after it, so it cannot cite a Pylint rule the way this guard polices; a blanket file-level suppression is a bigger problem that belongs to review, and none exist in this repo. That is now in the module docstring's Scope section.

**Gates:** `make check` clean. `make test` 11355 passed / 4 pre-existing skips. `tests/moneybin/test_architecture/` + `test_documentation_policy.py` 140 passed.

All three rounds have landed on the guard rather than the sweep, which seems right — the sweep is mechanical and provably comments-only, while the guard is the part actually making claims. The 83 swept Python files are untouched by rounds 2 and 3 and remain AST-identical to the base.

bsaffel:
Round 4 — both findings real, and the second is one I should have caught myself. Head `67ab81e2`.

**Root-level files were never scanned.** `SCANNED_TREES = ("src", "tests", "scripts")` skipped the tracked root `conftest.py`, which CI lints with `ruff check .`. The scan now asks the VCS for the actual tracked Python surface, and asserts the listing is plausibly large — an empty or failed listing must never read as "nothing to check", which would make the guard pass vacuously wherever that tool is unavailable.

**The scan never exercised its own predicate.** It only ever asserted an empty list, so swapping `MARKER` for a regex that never matches left the suite green. That is exactly the failure `.claude/rules/testing.md` names in *"A Fixture That Never Reaches the Predicate Proves Nothing"* — and I had cited that discipline in this PR's own narrative while shipping a guard that violated it. Now pinned by 9 positive cases covering every spelling it must catch (line-level, file-level, no-space, comma-joined, mixed case, PLW/PLR) and 6 negative ones covering the live suppressions it must not touch (`S404`, `S608`, `BLE001`, file-level `S101`, and prose that merely names a code).

**Verified by mutation, both directions:**

| Mutation | Before | After |
| --- | --- | --- |
| File-level marker planted in root `conftest.py` | invisible, green | ✅ red at `conftest.py:1` |
| `MARKER` replaced with a never-matching regex | green | ✅ 9 positive cases fail |

**One design note:** the cases interpolate the rule codes instead of spelling them out, so this file does not trip its own scan. The alternative was exempting the file from its own guard, and an exemption list is what I have been avoiding throughout — it is how the original 363 markers accumulated unnoticed.

**Gates:** `make check` clean. `make test` 11370 passed / 4 pre-existing skips — up 15, matching the new cases exactly. `tests/moneybin/test_architecture/` + `test_documentation_policy.py` 155 passed.

The swept 83 Python files remain untouched since round 1 and AST-identical to the base; rounds 2–4 are all confined to the guard.

PR #581, opened 2026-09-10
Honor configuration overrides while preserving profile defaults

bsaffel:
## Summary

Database, data, and logging settings supplied through environment variables or profile configuration were ignored because profile defaults were injected as higher-priority constructor values. Profile paths now come from a fallback settings source, so explicit constructor values, environment variables, and profile configuration take precedence while unspecified paths remain isolated to the selected profile.

The configuration reference now describes working overrides, and regression tests cover precedence, partial overrides, profile defaults, and settings-cache behavior.

## Test plan

- Configuration and reference regression tests: 66 passed, including failures reproduced before the fix.
- `make check`: passed.
- `make test`: 11,543 passed, 4 expected skips.
- `make test-integration`: 608 passed.
- `make test-scenarios`: 93 passed.
- `make test-e2e`: 380 passed; three wheel-packaging setups were blocked by sandbox DNS, then all three passed in the isolated network-enabled retry.


bsaffel:
@claude Please re-review the changelog citation finding at the current head, `5cd5b3f69fd80032e2b51d4935bf292daf8f7a58`, and reconcile review thread `PRRT_kwDOPjlNiM6g4y2s` if the fix satisfies it. `CHANGELOG.md:17` now reads `(PR #581)`; the placeholder was replaced in that commit. The documentation-policy gate passed all 43 tests after the change. The latest approval still describes the older placeholder. Please inspect the current file and provide the current-head disposition; no merge is requested.

PR #589, opened 2026-09-11
Populate manifest.provenance on bundle exports

### LINKED ISSUE #448: Bundle exports write manifest.provenance: null, so an artifact cannot name the build that made it
## What

A bundle export writes `manifest.provenance: null`. Report exports populate it. The artifact therefore carries no record of the build that produced it.

## Observed (live, v0.1.0 @ `e9c9ab15`, MCP)

```
export_run(subject={"kind":"bundle"}, destination={"kind":"local","name":"exports","format":"csv"},
           redaction_mode="redacted")
```

`manifest.json` in the emitted artifact:
```json
{
  "artifact_version": 1,
  "created_at": "2026-08-29T03:43:49.313486+00:00",
  "export_id": "878b0b6700944bd693952ebd25c6fa03",
  "profile": "brandon-release-1",
  "provenance": null,
  "redaction_mode": "redacted",
  "subject": {"kind": "bundle"}
}
```

Everything else in the receipt is solid — I re-hashed the files independently and all 8 declared checksums matched, and a second run produced byte-identical table checksums (only `manifest.json` differed). The one missing field is provenance.

## Mechanism

`src/moneybin/exports/snapshot.py:220` passes `provenance=None` for `ExportSubject(kind="bundle")`, and `snapshot.py:170` emits it verbatim. The report path populates it at `src/moneybin/exports/service.py:765`.

## Why it matters

The data exists and is already exposed elsewhere — `system_status` returns `overview.build` as `{"version": "0.1.0", "revision": "e9c9ab15a98a6925dee7941d2d0f0acf1e99a445"}`. A bundle is the **portability** artifact, the one most likely to outlive the machine that made it and to be read by another tool or a future version. Without a build stamp there is no way to answer "which code wrote this file" from the file itself, and no way to interpret a schema quirk against the version that produced it.

This also blocks release-gate evidence: a validation run is supposed to record the build behind each artifact, and the artifact cannot currently carry it.

## Done when

- A bundle export's `manifest.provenance` is populated on the same terms as a report export.
- It names at least the version and revision that `system_status.overview.build` reports.
- A test pins non-null provenance for both subject kinds.

## Notes

A prior validation catalog row (X.4) recorded this gap; this is an independent live reproduction on current `main`, so it is still open rather than fixed.

bsaffel:
Closes #448.

## Problem

A bundle export wrote `manifest.provenance: null` while a report export populated it. A bundle is the portability artifact — the one most likely to outlive the machine that made it — so it was the artifact least able to answer "which code wrote this file". The data already existed: `system_status.overview.build` reports the same version and revision.

## Change

The version/revision derivation moved out of the MCP adapter into a new `moneybin.build_info`, and both `system_status` and the export layer now read it from there. Putting it in a shared module rather than importing the adapter keeps the MCP→service layering direction intact; `mcp/tools/system.py` already carried a deferred import of `exports.service` to dodge a cycle, and reaching the other way would have tightened that knot.

`ReportExportProvenance` is generalized to `ExportProvenance`: a mandatory `build` for both subject kinds, with `report_id`/`receipt` now optional and populated for a report only. The alternative was a second, bundle-only provenance type sitting beside the report-only one — two shapes for one concept.

## Scope note

Two things go beyond the letter of the issue, both deliberate:

- **The report path also gained `build`.** The issue only asks for bundles, but a field that comes back null for one of two subject kinds sharing a type is a question a reviewer would have to ask. One shape, populated the same way for both.
- **`manifest.json` gains a key for report exports too.** This is an additive change to an on-disk format — a reader written against `artifact_version: 1` still finds `report_id` and `receipt` where it left them — so `ARTIFACT_VERSION` is unchanged. Flagging it because on-disk formats are a public contract; pre-launch posture in `design-principles.md` is what makes the additive change the right call rather than a version bump.

## Test plan

- `make check` — 0 errors (3 pre-existing `reportlab` warnings, unrelated).
- `make test` — 11561 passed, 4 skipped.
- `make test-integration` — 608 passed. Run because the diff changes an emitted artifact shape and touches the MCP surface.
- New: bundle provenance pinned non-null with `build` matching `get_build_info()`, and `report_id`/`receipt` null; the report test extended to pin the same `build`, so both subject kinds are covered per the issue's acceptance.

## Notes

Uses a `changelog.d/` fragment rather than a `CHANGELOG.md` edit, per #587.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,12 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 ## [Unreleased]
 
 ### Fixed
+- **A bundle export's manifest now names the build that produced it.**
+  `manifest.provenance` previously wrote `null` for a bundle export while a
+  report export populated it, leaving a bundle artifact with no record of
+  which code wrote it. Both subject kinds now carry the same version and
+  revision `system_status.overview.build` reports. (PR #TBD)
+
 - **Profile configuration now honors database, data, and logging overrides.**
   `MONEYBIN_DATABASE__*`, `MONEYBIN_DATA__*`, and `MONEYBIN_LOGGING__*` values
   from the environment or active profile dotenv now override profile defaults.
--- a/changelog.d/+bundle-export-provenance.fixed.md
+++ b/changelog.d/+bundle-export-provenance.fixed.md
@@ -1,0 +1,1 @@
+A bundle export's `manifest.provenance` now names the build that produced it instead of writing `null`, so an artifact can be interpreted against the code that wrote it after it leaves the machine. Report exports carry the same version and revision stamp that `system_status.overview.build` reports.
--- a/docs/specs/export.md
+++ b/docs/specs/export.md
@@ -82,10 +82,14 @@ format-neutral prepared snapshot. The snapshot contains:
 - the ordered data tables and typed columns;
 - the profile, UTC creation timestamp, export kind, and format version;
 - the selected redaction mode and output column classes;
-- row counts and per-table checksums; and
+- row counts and per-table checksums;
+- the build that produced it — package version and source revision, the same
+  pair `system_status.overview.build` reports; and
 - a generated data dictionary.
 
-For a report, the snapshot additionally carries its provenance receipt: report
+Both export subjects carry that build stamp — it is every artifact's only
+record of which code wrote it once the file leaves this machine. For a
+report, the snapshot additionally carries its provenance receipt: report
 identifier, resolved parameters, SQL, lineage, output classes, freshness,
 graduation eligibility, and classification-drift state. This is the same
 information exposed by the report verification surface, rendered as artifact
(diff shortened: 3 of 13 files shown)


bsaffel:
Fixed in `624bc1ad` — `docs/specs/export.md` now documents `provenance.build`.

**🟡 CONSIDER — `provenance.build` undocumented in an `implemented` spec:** addressed in both places the spec described provenance as report-only.

- **R2, snapshot contents (~line 84):** the build stamp is now listed as snapshot content, with an explicit statement that *both* export subjects carry it.
- **Artifact manifest (~line 446):** documents the field's shape, that either key is `null` when unresolvable, that it derives from the one shared `moneybin.build_info` module so a manifest and a running server can't disagree about which code produced them, and that a bundle manifest's `provenance` holds only `build` with `report_id`/`receipt` present as `null`.

This also answers @chatgpt-codex-connector's P1 on `src/moneybin/exports/snapshot.py:239`, which flagged the same report-only framing in the spec text.

Because this documents an on-disk contract, the two falsifiable claims in the new prose were checked against the code rather than asserted:

- `report_id`/`receipt` really are **present as `null`** for a bundle rather than omitted — the serializer emits all three keys unconditionally (`snapshot.py:314-316`). A verifier can therefore rely on the keys existing.
- Both `version` and `revision` are genuinely nullable on `BuildInfo`, and `read_git_revision` already documents the installed-wheel case as `null`.

**Gate run:** `uv run pytest tests/test_documentation_policy.py` → 52 passed. The diff is documentation-only (19 insertions, 4 deletions, one file), so no `.py` gate applies. `ReportExportProvenance` greps to zero hits tree-wide, so the rename left no orphaned prose.

PR #593, opened 2026-09-11
Resolve superseded transaction ids before curation writes

### LINKED ISSUE #538: Curation written under an already-superseded transaction id is unreachable forever
## The defect

`TransactionService.add_note()`, `add_tags()`, and `add_split()` accept a
caller-supplied `transaction_id` and write it straight through to the curation
table without checking that it names a transaction that exists.

Separately, the alias-forwarding derivation added in #406 permanently excludes
any id that already carries an alias — forwarding runs once, at the moment a
canonical id is superseded, and an id already on the left-hand side of the map
is never a forwarding candidate again.

Together those two facts make one class of curation invisible forever: curation
written under an id that was *already* superseded lands on a row no view joins,
and no later forwarding pass will ever pick it up. `core.fct_transactions` joins
`app.transaction_categories` / `_notes` / `_tags` / `_splits` on
`transaction_id` directly, so the note simply never appears.

This is **not** a regression from #406 — the write path accepted an unresolvable
id before that PR too, and #406 neither opens nor closes it. It was found while
adversarially reviewing that PR and is filed separately so it is not lost.

## Why the fix belongs at the service boundary

Guard 2 in `.claude/rules/identifiers.md` puts id resolution at the point of
entry, not at the point of read. Resolving through the alias map on read would
mean repeating that resolution in `core.fct_transactions`, in the doctor's FK
invariants, and in every curation repo — several mechanisms for one fact, with
the orphaned rows still sitting in the table meanwhile.

One seam covers all four curation tables: on write, resolve the supplied
`transaction_id` against `core.fct_transactions`; if absent, walk
`app.transaction_id_aliases` to its current canonical id; if that also fails,
raise `UserError` rather than writing an unreachable row.

## Acceptance

- A curation write against a superseded id resolves to the live transaction.
- A curation write against an id that names nothing is refused with a
  `UserError`, not silently accepted.
- One resolution seam, shared by notes, tags, splits, and categories.
- `system doctor`'s curation FK invariants stay green across a merge that
  re-keys an id a caller had just written to.

bsaffel:
Closes #538.

## Problem

`add_note`, `add_tags`, and `add_split` wrote a caller-supplied `transaction_id` straight to the curation table with no check that it named a live transaction. Alias-forwarding only re-forwards at the moment an id is superseded, so curation written under an *already*-superseded id landed on a row no view joins — `core.fct_transactions` joins the curation tables on `transaction_id` directly — and no later forwarding pass would ever pick it up. The note simply never appeared, permanently.

## Change

One write-time seam, `resolve_curation_transaction_id`, in `matching/aliasing.py`: try the id as given, walk `app.transaction_id_aliases` to the current canonical id if it names nothing, and raise `UserError` if neither resolves. Wired into every curation writer that accepts a bare id — notes, tags, splits, and categories — so the four paths share one mechanism.

Read-time resolution was the alternative and the issue rules it out: it would repeat the walk in `core.fct_transactions`, in the doctor's FK invariants, and in every curation repo, with the orphaned rows still sitting in the table meanwhile. The module docstring previously asserted that nothing resolves through the alias map; it now records this write-time exception rather than leaving the claim stale.

Liveness is tested against `core.fct_transactions` **or** `raw.manual_transactions` — the same pair the doctor's `orphan_app_state` invariant unions, verified against that code, so a manual entry written moments before its `refresh_run` doesn't misread as unresolvable.

## Behavior change — worth a reviewer's eye

This refuses writes that are accepted today, which is what the issue asks for, but it is a real tightening: a curation write that CREATES or CHANGES state (a note, a non-empty tag or split set, a category) whose id names nothing at all — not the id itself, and nothing it forwards to — now raises `transaction_reference_not_found` instead of silently producing an unreachable row.

A write that only REMOVES state is exempt by design, `clear_category` included: it never refuses for liveness and stays the idempotent no-op it always was, per the add-vs-remove asymmetry `resolve_curation_transaction_id`'s docstring describes (`test_clear_category_against_unknown_id_stays_a_noop` pins this). An earlier draft of this PR description claimed `clear_category` had started refusing on an unknown id — it hadn't; that line was wrong and is corrected here.

This is recorded in a `changed` changelog fragment alongside the `fixed` one.

`TestSetCategoryAudit` also needed a seeded transaction — it categorizes a bare `"T1"` that exists in no relation, so the new seam refused before any audit row was written. Seeding the fixture was the right fix there rather than weakening the seam; those tests assert audit emission, not id resolution.

## Test plan

- `make check` — 0 errors (3 pre-existing `reportlab` warnings, unrelated).
- `make test` — 11574 passed, 4 skipped.
- `make test-integration` — 609 passed. Run because the diff touches the matching/aliasing primitive.
- `make test-scenarios` not run locally; CI's scenario shard covers it.
- New: curation under a superseded id resolves to the live transaction; an unresolvable id refuses; and the doctor's `app_transaction_categories_fk` invariant is asserted `pass` across the re-key, which is the issue's own acceptance criterion.

All fixtures synthetic; no live profile read.

## Notes

Uses `changelog.d/` fragments rather than a `CHANGELOG.md` edit, per #587.


bsaffel:
Both 🔴 MUST FIX items were already resolved as of `9f8fdb82` (verified, not redone) — pushed now as `8f7e4e19` alongside the blocking design fix below.

**1. Guarded categorization write path.** `write_categorization`/`write_categorizations` (`applier.py:2072`, `:2132`) resolve the caller-supplied id through the shared seam before writing and refuse `transaction_reference_not_found` when unresolvable. The three other internal callers of the shared guarded write — `apply_rules`, `apply_merchant_categories` (`orchestrator.py:710`, `:880`), and `auto_rule_service.py`'s rule-approval backfill (`:1414`) — all route through `write_categorization(s)` and inherit the seam for free. The two loop callers (`orchestrator.py:280`, `auto_rule_service.py:1395`) bulk-resolve once ahead of their loop via `resolve_curation_transaction_ids` and pass `resolve_transaction_id=False`, so the loop stays O(1) queries rather than one per row — the measured cost of the per-row alternative is ~319x slower at 5,000 rows.

**2. `transactions_annotate`.** `TransactionService._prepare_annotation` resolves `transaction_id` for `NoteAdd`/`TagsSet`/`SplitsSet` before validating or diffing, and `_apply_prepared_annotation` writes through the resolved id (the prepared target/mutation), never the caller's raw one.

## The add-vs-remove asymmetry

`make test` on `9f8fdb82` showed 10 failures. One — `test_apply_annotations_clears_existing_orphan_note_and_tags` — exists specifically to clear state on a dead id: orphan cleanup. A uniform "always refuse on liveness" rule breaks that test's entire purpose, and the reviewer's 🟡 flagged the same tension from the other side (`remove_tags`/`clear_splits` losing their old idempotent-no-op behavior).

Rule implemented:

- A write that **creates or changes state** — add a note, add/set tags with a non-empty desired list, add/set splits with a non-empty desired list, set a category — resolves through aliases and refuses `transaction_reference_not_found` when unresolvable.
- A write that **only removes state** — remove tags, clear splits, clear a category, or set an empty tag/split list — never refuses for liveness. It resolves to the live id when one exists, and otherwise falls through to operate on the id as given: the original idempotent no-op (DN2).

Implemented once, in `resolve_curation_transaction_id`'s new `required: bool = True` parameter (`aliasing.py`) — not as scattered per-call-site special cases. Every pure-removal writer (`clear_category`, `remove_tags`, `clear_splits`) passes `required=False` unconditionally. Every "set"-shaped writer (`set_tags`, `set_splits`, and their `TagsSet`/`SplitsSet` batch-annotate counterparts) computes `required` from whether the caller's own desired state is non-empty — a static predicate of the request, not the realized diff (deriving it from the diff would need the resolution outcome to already exist). Why removals must stay permissive: refusing them would permanently block cleanup of curation stranded on an id no automatic healing pass can reach — exactly the scenario the orphan-cleanup test pins.

## 🟡 #3 — new coverage

Added direct tests proving both halves of the asymmetry per writer:

- Permissive on an unknown id: `test_remove_tags_against_unresolvable_id_stays_a_noop`, `test_clear_splits_against_unresolvable_id_stays_a_noop`, `test_clear_category_against_unknown_id_stays_a_noop`.
- `set_tags`/`set_splits` still refuse when the desired list is non-empty: `test_set_tags_against_unresolvable_id_raises_user_error`, `test_set_splits_against_unresolvable_id_raises_user_error` — but resolve permissively when clearing: `test_set_tags_clear_against_unresolvable_id_stays_a_noop`, `test_set_splits_clear_against_unresolvable_id_stays_a_noop`.
- `test_apply_annotations_clears_existing_orphan_splits` mirrors the existing note/tag orphan test so the batch `transactions_annotate` surface pins the same asymmetry for splits.

Also reverted two tests from an earlier commit on this branch that had pinned the old (pre-ruling) strict behavior for `clear_category`, and updated the changelog fragment plus two query-count assertions to account for the new resolve call the seam now performs.

## Disposition of the 10 failures

- 1 (`test_apply_annotations_clears_existing_orphan_note_and_tags`) — fixed by the asymmetry itself; it was never a fixture gap.
- 7 (Plaid-categorizer and improve-AI tests across three test files) — fixture-seeding gaps: these tests exercise the categorizer against `prep.int_transactions__merged` only, with no matching `core.fct_transactions` row, so the newly-added resolution correctly refused a truly-nonexistent gold id. Seeded the missing row per test (same fix pattern as the earlier `ddba9072` commit on this branch).
- 2 (`test_categorize_items_uses_constant_number_of_db_calls`, `test_batches_precedence_skip_diagnostics`) — query-count assertions that needed to account for the new O(1) bulk-liveness query the seam adds; both now assert on the distinct query shapes rather than one combined count.

## Gates run

- `make check` — clean (format/lint/pyright), twice (once before, once after the final round of edits).
- `uv run pytest` across 25 targeted files (alias healing, categorization service + writes + matcher + shared + rule-conflicts + updated_at, transaction service, auto-rule service, MCP categorization/curation/transactions/review tools, and the categorize CLI command group) — **759 passed, 0 failed**, 232.7s.

Full test-file list for the integration/scenario suites you're running separately, all under the same command: `tests/moneybin/matching/test_transaction_id_alias_healing.py`, `tests/moneybin/test_services/test_categorization_service.py`, `tests/moneybin/test_services/test_categorization_service_writes.py`, `tests/moneybin/test_services/test_transaction_service.py`, `tests/moneybin/test_mcp/test_categorization_tools.py`, `tests/moneybin/test_mcp/test_curation_tools.py`, `tests/moneybin/test_cli/test_cli_transactions_tags.py`, `tests/moneybin/test_cli/test_cli_transactions_splits.py`, `tests/moneybin/test_cli/test_cli_transactions_notes.py`, `tests/moneybin/test_fct_transactions_curation.py`, `tests/moneybin/test_cli/test_cli_transactions_audit.py`, `tests/moneybin/test_mcp/test_transactions_tools.py`, `tests/moneybin/test_mcp/test_transactions_get.py`, `tests/moneybin/test_categorization_service_items.py`, `tests/moneybin/test_services/test_auto_rule_service.py`, `tests/moneybin/test_services/test_categorization_updated_at.py`, `tests/moneybin/test_services/test_categorization_matcher.py`, `tests/moneybin/test_services/test_categorization_rule_conflicts.py`, `tests/moneybin/test_services/test_categorization_shared.py`, `tests/moneybin/test_cli/test_categorize_auto_commands.py`, `tests/moneybin/test_cli/test_categorize_commit_commands.py`, `tests/moneybin/test_cli/test_categorize_pending_command.py`, `tests/moneybin/test_cli/test_categorize_rules_commands.py`, `tests/moneybin/test_auto_rule_service_context.py`, `tests/moneybin/test_mcp/test_review.py`.

Did not run `make test`, `make test-integration`, or `make test-scenarios` (each exceeds this session's command timeout) — those are yours to run per the instructions above.

bsaffel:
## Root cause

`_live_transaction_ids` (used by both `resolve_curation_transaction_id` and
`resolve_curation_transaction_ids`) unions `core.fct_transactions` with
`raw.manual_transactions` to decide liveness, without checking that
`core.fct_transactions` is actually in the catalog first. `core.fct_transactions`
is a SQLMesh view built by the transform, so any DB before the transform has run
(a first load) has `app`/`raw` schemas but no fact view yet — every symbol on this
resolution path is new in this PR, so this is a genuine regression from it.

`tests/integration/test_categorize_improve_ai_cli.py`'s two failing tests seed
only a bridge mapping plus one `app.transaction_categories` row and never run the
transform, so the union query threw `CatalogException`, which propagated as a
CLI exit code of 1.

## Ruling

Absence of the liveness oracle is absence of evidence, not evidence the id is
dead. When `core.fct_transactions` isn't in the catalog, resolution now passes
the id through unchanged — it must not crash, and it must not raise
`TRANSACTION_REFERENCE_NOT_FOUND` either, since that would also red these tests
(just with a nicer message) and would newly break every pre-transform curation
write that worked before this PR.

This mirrors `_heal_stranded_curation`, which already takes exactly this posture
for the same missing relation (`if not _relations_exist(db, FCT_TRANSACTIONS,
TRANSACTION_ID_ALIASES): ... return 0`) with the same reasoning in its comment:
"A first load precedes the transform that builds the fact view." Fail-closed
buys no integrity here — the doctor's FK invariants plus `_heal_stranded_curation`
are the existing backstop for curation attached to an id no view serves.

## Fix

Both `resolve_curation_transaction_id` (single-id, both `required=True` and
`required=False`) and `resolve_curation_transaction_ids` (bulk) now probe
`_relations_exist(db, FCT_TRANSACTIONS, MANUAL_TRANSACTIONS)` once, up front,
before touching the liveness query — reusing the existing catalog-probe idiom
rather than adding a second one, and never wrapping the query in a
`try`/`except CatalogException` (which would leave a caller's transaction
poisoned, per the module's own documented reasoning). If the oracle is absent,
the id (or every id, for the bulk path) is returned unchanged with a debug log,
and the bulk path short-circuits before ever reaching its per-id fallback loop —
so a first-load batch write costs one extra catalog lookup total, not one per row.

## Mutation proof

Removed both guards, ran the three new unit tests
(`tests/moneybin/matching/test_transaction_id_aliasing.py::TestResolutionBeforeTheFactViewExists`)
against a DB with `core.fct_transactions` dropped:

```
FAILED ...test_single_id_required_true_passes_through_unchanged
FAILED ...test_single_id_required_false_passes_through_unchanged
FAILED ...test_bulk_resolution_returns_an_identity_map_for_every_id
_duckdb.CatalogException: Catalog Error: Table with name fct_transactions does not exist!
3 failed in 4.12s
```

Restored the guards — file diffed identical to the committed version — and the
same three tests passed in 3.75s.

## Gates run

- `make check` (repo-wide) — 0 errors, 3 pre-existing unrelated pyright warnings.
- `uv run pytest tests/integration/test_categorize_improve_ai_cli.py -m integration` — 3 passed.
- `tests/moneybin/matching/test_transaction_id_aliasing.py`,
  `tests/moneybin/matching/test_transaction_id_alias_healing.py`,
  `tests/moneybin/test_services/test_transaction_service.py`, and the
  categorization service test files (`test_categorization_rule_conflicts.py`,
  `test_categorization_service_writes.py`, `test_categorization_shared.py`,
  `test_categorization_service.py`, `test_categorization_matcher.py`,
  `test_categorization_updated_at.py`, `test_categorization_audit.py`) — 462 passed.

One incidental test needed updating:
`test_categorization_service_writes.py::TestBatchCategorizationWrites::test_batches_precedence_skip_diagnostics`
pinned an exact DB-call count for the bulk write path; the new catalog probe adds
one call, so the count and its comment moved from 5 to 6.

bsaffel:
Addressed all 7 findings from the latest review round (4 🔴, 2 🟡, 1 🔵) on top of `5f794082`. New tip: `81192a62`.

## 🔴 1 — deduped-away manual liveness gate (`matching/aliasing.py`)

Fixed by excluding a `raw.manual_transactions` row with an outgoing `app.transaction_id_aliases` edge from the liveness union — `_live_transaction_ids`'s manual arm now does `NOT EXISTS (SELECT 1 FROM app.transaction_id_aliases WHERE old_transaction_id = m.transaction_id)`. A fresh, not-yet-aliased manual row is untouched (no edge to exclude).

**Ordering/probe decision (as requested):** kept liveness-first ordering — no per-id alias round trip added. `app.transaction_id_aliases` is **not** added to the `_relations_exist` probe: it's a plain `app.*` schema table created at DB init (unlike the transform-produced `core.fct_transactions`/`raw.manual_transactions`), so it always exists whenever this query runs at all. The probe only needs to guard the two relations a first load genuinely precedes. Pre-transform pass-through tests (`TestResolutionBeforeTheFactViewExists`) pass unchanged.

**Mutation-proof:** reverted the anti-join, re-ran the new tests — `assert 'manual_abc' == 'csv_xyz'` → `AssertionError: assert 'manual_abc' == 'csv_xyz'` (single-id) and `assert {'manual_def': 'manual_def'} == {'manual_def': 'csv_uvw'}` (bulk). Restored, both green.

## 🟡 6 — bulk fallback repeats the reversed-edge scan (`aliasing.py`)

Split the hop/cycle walk out of `resolve_curation_transaction_id` into `_walk_alias_chain(db, id, *, required, reversed_edges=None)`. `resolve_curation_transaction_ids`'s fallback loop now computes `_reversed_alias_edges` once for the whole stale-id set and threads it through, instead of each fallback id recomputing its own (also skips the redundant `_relations_exist` re-probe, already known true).

**Mutation-proof:** reverted the fallback loop to call the old per-id `resolve_curation_transaction_id`, ran a new query-count test with 2 stale ids sharing one live target — `AssertionError: Expected exactly 1 reversed-edge scan for 2 stale ids in one batch, got 2`. Restored, green.

Also fixed the 🔵 NIT (malformed single-backtick closing `` `execute()` `` in both `aliasing.py` and `applier.py`).

## 🔴 2 — `_reject_composed_annotations` keys on the unresolved id (`transaction_service.py`)

Verified the claim by tracing the `else` branch: `key = (request.kind, request.transaction_id)` used the caller's raw id. Changed to `key = (request.kind, item.target_ids[0])` — the id already resolved a few lines earlier.

**Mutation-proof:** new test seeds `T1` with a split, sends two `SplitsSet` clears — one against `T1_OLD` (aliases to `T1`), one against `T1` directly. Reverted the fix: `Failed: DID NOT RAISE <class 'moneybin.errors.UserError'>` — both applied silently in sequence. Restored: raises `mutation_invalid_input` as expected, `list_splits("T1")` still has its 1 original row.

## 🔴 4 — `TagsSet` permissive clear still raises on a dead id (`transaction_service.py`)

Confirmed by trace: `if request.tags or not mutation.to_remove:` evaluated `False or True → True` for an empty-tags request against a dead id with nothing existing to remove, so `_annotation_transaction_amount` still ran and raised. Changed to `if request.tags:`, mirroring the `SplitsSet` branch immediately below.

**Mutation-proof:** new test batches `NoteAdd(T1, ...)` with `TagsSet(NEVER_EXISTED, tags=[])`. Reverted: `moneybin.errors.UserError: The transaction reference did not match a transaction.` (raised from `_annotation_transaction_amount`), aborting the whole batch including the unrelated `NoteAdd`. Restored: batch succeeds, `outcomes == [True, False]`.

## 🔴 3 — no resolved-id collision check in the categorize batch (`categorization/orchestrator.py`)

Verified the claim, then mirrored `_reject_composed_annotations`'s shape rather than inventing a new pattern: a `Counter` over `resolved_txn_ids` values, checked per-item in Phase 4; every item whose resolved id collides is rejected (not just the second) with `error: "resolved_id_collision"`, keeping this batch's existing per-item soft-failure design (unlike the annotate path's whole-batch hard raise — `transactions_categorize_commit` is documented as per-item-failure-tolerant, so a batch-level raise would be the wrong shape here).

**Mutation-proof:** new test sends a superseded id and its canonical id in one batch, each with a different category. Reverted the check: `assert 2 == 0` — `CategorizationResult(applied=2, skipped=0, errors=0, ..., merchants_created=1)` — both silently applied, exactly the "second wins, both counted" bug described. Restored: `applied=0, errors=2`, no row written.

## 🟡 5 — Phase 2/4 key on the pre-resolution id (`orchestrator.py`)

Fixed rather than deferred, per your instruction. Phase 2 now fetches `core.fct_transactions` rows by `resolved_txn_ids.values()` instead of the caller's original ids; Phase 4's `ctx.description_for`/`memo_for`/`merchant_name_for`/`merchant_entity_id_for`/`merchant_entity_source_type_for` and `auto_rule_svc.record_categorization` all use `resolved_txn_id`.

**Mutation-proof:** new test categorizes a superseded id with `canonical_merchant_name="Starbucks"` set on a transaction whose description is `STARBUCKS COFFEE`, and asserts `merchants_created == 1` plus a non-empty exemplar list in `app.user_merchants`. Reverted just the Phase 2 fetch (`fetch_ids` → `txn_ids`): `assert 0 == 1` — `CategorizationResult(applied=1, ..., merchants_created=0)`. Restored: `merchants_created == 1`, exemplar row present.

## Gates run

- `make check` (format + lint + pyright), repo-wide bare run: **0 errors** (3 pre-existing warnings in an unrelated `reportlab`-import fixture-generation script, untouched by this diff).
- Targeted tests, foreground, real summary line read directly (not piped through `tail`):
  - `tests/moneybin/matching/test_transaction_id_aliasing.py` + `test_transaction_id_alias_healing.py` + `test_services/test_transaction_service.py` + `test_services/test_categorization_service.py` → **290 passed**
  - Wider blast-radius sweep (`test_recording_context.py`, `test_categorization_service_items.py`, `test_exemplar_accumulator.py`, `test_auto_rule_service.py`, `test_source_precedence.py`, `test_categorization_audit.py`, `test_categorization_rule_conflicts.py`, `test_categorization_updated_at.py`, `test_categorization_service_writes.py`, `test_transaction_get.py`, `test_matching_service.py`) → **305 passed**
  - MCP/CLI annotate + categorize surfaces (`test_categorization_tools.py`, `test_transactions_tools.py`, `test_categorize_commit_commands.py`, `test_cli_transactions_notes/splits/tags.py`) → **94 passed**
  - Pre-existing query-count contract test (`test_categorize_items_uses_constant_number_of_db_calls`) still green — my batching change to the aliasing fallback doesn't shift it (that test has no stale ids to exercise the fallback path).

**I did not run `make test` / `make test-integration` / `make test-scenarios`** — the dispatching session is running the full unit/integration/scenario suites separately on this pushed tip.

4 commits on top of `5f794082`: `7e2c66f0` (🔴1 + 🟡6 + NIT in aliasing.py), `e2a23b72` (🔴2 + 🔴4 in transaction_service.py), `e89fc9f1` (🔴3 + 🟡5 in orchestrator.py), `81192a62` (NIT in applier.py).

bsaffel:
Addressed all 6 items from this review round (5 🟡, 1 🔵) on top of `49fa0fc0`. New tip: `1237ba15`.

## 🟡 1 — collision guard regressed the verbatim-duplicate case (`orchestrator.py:287`)

Confirmed on both of us. Fixed by keying the collision check on the count of **distinct raw ids** resolving to a live transaction, not raw occurrence count — `resolved_id_sources: dict[str, set[str]]` instead of a `Counter`. A literal duplicate `transaction_id` submitted twice now falls through to `write_categorization`'s pre-existing source-priority guard, unchanged.

**Mutation-proof:** added `test_categorize_items_verbatim_duplicate_keeps_priority_guard_behavior` (two items, same live id, no alias). Reverted to the old occurrence-count logic: `assert 2 == 0` → `AssertionError: assert 0 == 2` — `CategorizationResult(applied=0, errors=2, ..., 'error': 'resolved_id_collision')`, i.e. the exact regression. Restored: `applied=2, errors=0`, second same-priority write overwrites the first (verified this really is `<=` in `write_categorization`'s guard before writing the test, not the `lower_priority_source`-skip I initially assumed). The pre-existing aliased-duplicate test (`test_categorize_items_rejects_two_ids_resolving_to_the_same_live_transaction`) still passes unchanged.

## 🟡 2 — forward-only walk vs. the healer's undirected walk (`aliasing.py`)

Fixed by reusing `_heal_stranded_curation`'s own undirected-component query rather than writing a second walk (design-principles coherence rule). Refactored `_STRANDED_CURATION_SQL` into two pieces: `_STRANDED_CURATION_IDS_SQL` (which curation ids are stranded) and `_LIVE_COMPONENT_SQL` (the shared undirected-component-with-live-count query, now parameterized by a seed-id list instead of hardcoding the stranded-curation seed). `_heal_stranded_curation` calls it seeded from stranded curation ids (unchanged behavior — full healing test suite still green); a new `_live_predecessor(db, transaction_id, reversed_edges)` calls the same query seeded from one caller id, invoked by `_walk_alias_chain` once the forward walk is exhausted.

**Ambiguity decision:** when the undirected component holds more than one live member, `_live_predecessor` returns `None` — treated identically to "unresolved" (raises for `required=True`, passes through unchanged for `required=False`). Never guesses.

**Mutation-proof:**
- Positive case (`test_a_caller_held_anchor_id_resolves_to_the_live_predecessor`, via the real `revert_confirmed` mechanism): reverted the predecessor-fallback call → `moneybin.errors.UserError: The transaction reference did not match a transaction.` Restored: resolves to the live predecessor.
- Ambiguity case (`test_an_ambiguous_predecessor_component_is_refused_not_guessed`): reverted just the `live_count != 1` guard → `Failed: DID NOT RAISE`. Restored: raises `transaction_reference_not_found`.
- Full `test_transaction_id_alias_healing.py` (48 tests incl. the pre-existing `TestAnUndoneMergeStopsTheWalk`/reversed-edge regression tests) still green — the refactor didn't change `_heal_stranded_curation`'s observable behavior.

## 🟡 3 — two liveness definitions (`transaction_service.py:496`)

Verified the granular-writer claim directly: `add_note` never calls `_annotation_transaction_amount` at all (it doesn't need the amount), so it never hit this. Fixed `_annotation_transaction_amount` to fall back to `raw.manual_transactions` when the id has no row in `core.fct_transactions` — by the time an id reaches this call, `resolve_curation_transaction_id` has already forwarded any aliased-away manual id onward, so one still self-naming and missing from the fact view can only be an un-aliased manual row; no separate alias re-check needed.

**Mutation-proof:** new test creates a manual transaction via `create_manual_batch`, confirms it's absent from `core.fct_transactions`, then adds a note via `apply_annotations`. Reverted the fallback → `UserError: The transaction reference did not match a transaction.` at `_annotation_transaction_amount`. Restored: note lands.

## 🟡 4 — per-id resolution in the annotate batch (`transaction_service.py:495`)

Added `_bulk_resolve_annotation_ids` (one `resolve_curation_transaction_ids` call per `preview_annotations` invocation) and `_resolve_annotation_transaction_id` (applies each request's own `required` semantics against the batch's resolved-id map — permissive clears stay permissive, non-empty adds still raise). `_prepare_annotation` now takes the resolved map instead of calling the single-id resolver.

**Mutation-proof:** new query-count test (5 `NoteAdd`s in one batch) asserts exactly 1 bulk liveness query. Reverted `NoteAdd`'s branch back to the single-id call → `AssertionError: ... got 6` (1 bulk + 5 individual). Restored: 1. Full `test_transaction_service.py` (117 tests) stays green, confirming the permissive/strict semantics didn't shift.

## 🟡 5 — `create_manual_batch` resolves ids it just minted (`applier.py:478`)

Confirmed from `ManualEntryRawResult`'s own docstring and the raw-insert flow: `transaction_id` is the SHA256 gold key this exact call predicted moments earlier, so it cannot be aliased yet. Added `resolve_transaction_id: bool = True` to `set_category_in_active_txn` (mirroring `write_categorization`'s existing fast path), threaded through the `CategorizationService` facade; `create_manual_batch`'s loop now passes `resolve_transaction_id=False`.

**Mutation-proof:** new query-count test (3 categorized manual entries) asserts zero transaction-id-resolution queries. Reverted the call site to omit the flag → `got 3`. Restored: `0`. Categorization still lands correctly (`cat_count == (3,)`).

## 🔵 NIT — PR description contradicted shipped behavior

Corrected. The "Behavior change" section claimed `clear_category` against an unknown id now refuses — it doesn't; `clear_category` resolves with `required=False` and stays the idempotent no-op it always was (`test_clear_category_against_unknown_id_stays_a_noop` pins this). Rewrote that section to state the real, single behavior change (a create-or-change write refuses; a remove-only write, `clear_category` included, never does) and to say plainly that the earlier claim was wrong.

## Gates

- `make check` (format + lint + pyright), repo-wide bare run: **0 errors** (same 3 pre-existing, unrelated `reportlab`-import warnings).
- Targeted suites, foreground, real summary lines read directly (never piped through `tail` for the exit code):
  - `test_transaction_id_aliasing.py` + `test_transaction_id_alias_healing.py` → **48 passed**
  - `test_transaction_service.py` → **117 passed**
  - `test_categorization_service.py` → **131 passed**
  - `test_categorization_service_writes.py` → **101 passed**
  - `test_mcp/test_categorization_tools.py` + `test_mcp/test_transactions_tools.py` → **68 passed**
  - `test_cli_transactions_notes/splits/tags.py` + `test_categorize_commit_commands.py` → **26 passed**
  - Wider blast-radius sweep (`test_recording_context.py`, `test_exemplar_accumulator.py`, `test_auto_rule_service.py`, `test_source_precedence.py`, `test_matching_service.py`) → **100 passed**
- Ran everything with `-n2` (not `-n auto`) per the memory-pressure note.
- **Did not run** `make test` / `make test-integration` / `make test-scenarios` — you've already run those on this tip (unit 11607, integration 609, scenarios 93 on `49fa0fc0`; I did not re-trigger them).

No deferral issues filed — all 6 items were fixed in code this round, none deferred.

3 commits on top of `49fa0fc0`: `d10a6d44` (🟡1, orchestrator.py), `25ac2c7a` (🟡2, aliasing.py), `1237ba15` (🟡3+🟡4+🟡5, transaction_service.py/applier.py/categorization facade — grouped because all three land in the same annotate/manual-batch seam). One correction: `25ac2c7a`'s message describes a docstring fix that actually shipped in the prior commit (`49fa0fc0`), not this one — noted here since I caught it only after committing and didn't want to amend a pushed commit; the code itself is unaffected.

PR #600, opened 2026-09-12
Reconcile Plaid sync specifications with the shipped client

bsaffel:
## Summary

Make [the Plaid sync specification](https://github.com/bsaffel/moneybin/blob/codex/mb-72-plaid-spec/docs/specs/sync-plaid.md) the indexed canonical record and move the earlier design document into the archive with a short pointer to its replacement. The resulting documentation describes the shipped client boundary: Link initiation uses `POST /sync/link/initiate`, status reads use `GET /sync/link/status`, text CLI linking may wait for completion, and JSON CLI and MCP callers use later status reads.

It also reconciles the sync lifecycle with the implemented client: the trigger request waits for server completion before the client retrieves and loads the payload, connection health comes from the server-owned `GET /institutions` state, and no local `app.sync_connections` mirror is claimed. The Plaid extractor and MCP tool docstrings now point readers to the canonical specification, and the spec retains the raw-to-staging sign-convention boundary.

No changelog fragment is included because this is an internal documentation and specification alignment with no user-observable behavior change.

## Test plan

- `uv run pytest tests/test_documentation_policy.py -v` — 52 passed after the final documentation repairs.
- `git diff --check` — clean after the final documentation repairs.
- Historical MCP documentation corpus — 50 passed after moving the archived reference.
- `make check` — passed (Ruff, SQL formatting, and Pyright; Pyright reported only the repository's existing three missing-source warnings for reportlab).
- `PYTEST_ADDOPTS='-n 2' UV_CACHE_DIR=/private/tmp/mb72-uv-cache TMPDIR=/private/tmp/mb72-pytest-tmp make test` — 11,598 passed, 2 skipped. This ran before the final documentation-only correction; no implementation code changed afterward.


bsaffel:
Addressed all four review findings in the local repair being pushed next:

- Moved the superseded design into `docs/specs/archived/` with `Status: archived` and corrected its references.
- Corrected the trigger response to `job_id`, `status`, and `transaction_count`; per-institution outcomes come from `GET /sync/data` at `metadata.institutions`.
- Placed best-effort acknowledgement after durable raw loading and before transforms, including loss-free re-pull behavior when acknowledgement fails.
- Removed unavailable job-ID recovery after a trigger timeout. The guidance now distinguishes retrying the pull from inspecting connection health.

Validation: `make check` passed; documentation policy 52 passed; historical documentation corpus 50 passed; diff check clean.

bsaffel:
The next repair replaces the obsolete temporary-file recipe with the shipped `PlaidExtractor` path: typed sync models become Polars DataFrames and enter through `Database.ingest_dataframe()`. Payloads remain in memory. The adjacent removal recipe now describes the actual pre-count-then-delete behavior under the database write lock.

The investments and Google Sheets specs now agree that connection health is server-owned through `GET /institutions`. The data response carries one `status`, `error`, and `error_code` per institution at `metadata.institutions`; the docs explicitly identify product-specific error detail as unimplemented instead of promising an unsupported response shape or local table. The implemented investments child is also removed from future-work declarations.

Validation: documentation policy 52 passed; historical documentation corpus 50 passed; diff check clean. The changes are confined to four documentation specs.

bsaffel:
Updated the embedded accounts DDL to the shipped schema, including `persistent_account_id` and `name`, and corrected relink semantics: Plaid reissues `account_id`, while `persistent_account_id` is the cross-relink reference when available.

The error table now distinguishes pull output from connection-status guidance: pull reports raw per-institution codes without automatic retry or backoff, while `sync status` maps known health codes from `GET /institutions`. Documentation policy and the historical documentation corpus passed (52 and 50 tests). The branch also includes the latest bundle-export base change.

bsaffel:
Corrected Link-session follow-up to `sync_link_status(session_id=...)` in the three related specifications; `sync_status()` remains the no-argument connection-health read.

The overview now distinguishes raw per-institution pull outcome codes from `/sync/data` from mapped `/institutions` status guidance. Its provider contract uses the shipped extractor package layout. The Plaid spec links to the colocated DDL and staging views instead of maintaining duplicate SQL, while retaining the account-link, source-key, relink-identity, conversion-leg, and margin-balance semantics. The archive includes a verified, pinned Git retrieval command.

Documentation policy passed 52 tests and the historical documentation corpus passed 50. These changes address the three open consistency findings and the two occurrences of the incorrect MCP tool name.

bsaffel:
Correction to the prior remediation: I followed the internal helper name rather than the registered MCP surface. `register_sync_workflow_tools()` exposes `sync_status`, backed by `sync_status_coarse`; its registration test verifies the optional `session_id` parameter and explicitly verifies that `sync_link_status` is absent from the public tool names. All three specifications now use `sync_status(session_id=...)`.

The conversion-leg description now distinguishes reserved DDL columns from loaded data: `SyncTransaction` and the extractor schema currently omit `to_amount` and `to_currency`, so this ingestion path does not populate them.

The branch also includes the latest main commit. Post-sync verification passed: 9 registration tests, 52 documentation-policy tests, and 50 historical documentation-corpus checks.

The failed unit shard was the public MCP documentation contract test, which rejected those same unregistered tool names. Its setup succeeded; the shard otherwise had 2,585 passes and 2 skips. The corrected public-name references address that failure without changing the test or runtime surface.

Current-head verification also passed the exact formerly failing node (1 test) and the full MCP documentation-contract file (216 tests).

bsaffel:
The provider contract now distinguishes necessary provider artifacts from a working second-provider integration: current `get_data()` validates the Plaid-shaped response and both public builders instantiate `PlaidExtractor`, so the documented response-model and dispatch decision remains a prerequisite for provider two.

The investments record now names the shipped `PlaidExtractor` that handles its optional arrays. The Plaid test instructions now point to the existing SQL test and YAML fixture. Documentation policy passed 52 tests, and the full MCP documentation-contract suite passed 216 tests.

PR #601, opened 2026-09-12
Scope encrypted CLI initialization to its child process

bsaffel:
## Summary

Scope encrypted DuckDB CLI initialization to its child process while preserving interactive streams, temporary-script cleanup, and existing error guidance. This applies consistently to the shell, query, and UI entry points. Includes a security changelog fragment.

## Test plan

- Current head `322dc857d1e33260c272861890866562e441fb50`: focused CLI tests (61 passed), documentation-policy tests (52 passed), and `make check` passed with 0 Pyright errors; 3 existing optional dependency source-resolution warnings remain.
- Affected CLI, banner, and integration tests (`uv run pytest`, two workers): 69 passed after review repairs, including the isolated format-order case and installed-CLI environment lookup.
- E2E run: 380 cases passed. The three wheel-packaging setup errors were caused by a transient dependency-resolution DNS failure; an authorized isolated wheel build and the three wheel-packaging assertions then passed.

The E2E run covers `6a742574`; affected tests and quality checks also cover the review repairs. Current-head CI and external review remain required.


bsaffel:
The local repair being pushed next addresses the active error and test boundaries: all five key-error tests now raise from `_duckdb_cli_environment`, direct tests verify cleanup for both key-error types, and the format-order integration test and banner fixture provide the child environment. The obsolete error handlers around script creation are removed. All 69 affected tests pass, including the isolated integration case; formatting, lint, and repo-wide Pyright pass.

The `getenv` finding tested a different execution surface. DuckDB explicitly documents [`getenv` as CLI-only](https://duckdb.org/docs/current/clients/cli/overview), unsupported by its other clients. These commands invoke the DuckDB CLI, not the Python engine. The real CLI initialization path was exercised with synthetic data, and this repair adds an installed-CLI regression for environment lookup; it passed with CLI v1.5.5.

For the ATTACH helper suggestion, I retained the narrow CLI-specific statement. `build_attach_sql()` supplies literal keys to the Python engine, where `getenv` is unavailable; accepting arbitrary SQL expressions there would broaden that helper's contract. The CLI's fixed alias remains consistent with its existing `USE moneybin` statement. No shared database-helper API change is needed for this delivery.

bsaffel:
Updated the security specification and database-security guide to describe the current CLI contract: the init script contains a `getenv` lookup, the child environment supplies the value, and cleanup also covers environment-preparation failures. Removed the obsolete init-script test mock and classified the real installed-CLI probe as an integration test. The focused CLI module passed 61 tests, documentation policy passed 52, and `make check` passed.

The shared ATTACH helper remains unchanged for the narrow execution-boundary reason discussed above; the current review explicitly classifies that duplication as non-blocking. The branch also includes the latest bundle-export base change.

